PR #1, opened 2021-09-22
Some extra validation and allow to execute module with -m

Carreau:


tlambert03 (approved):
🚀

Carreau:
Just as a note, i'm not a superfan of using regex as then error messages are:

```
contributes -> commands -> 3 -> command
  string does not match regex "^((([a-zA-Z_][a-zA-Z_0-9]+)\.)*([a-zA-Z_][a-zA-Z_0-9]+))$" (type=value_error.str.regex; pattern=^((([a-zA-Z_][a-zA-Z_0-9]+)\.)*([a-zA-Z_][a-zA-Z_0-9]+))$)
```

Which is not great.

PR #3, opened 2021-09-27
Add support for writer plugins

nclack:
This adds the WriterContribution.

# Related
napari/napari#3426
napari/napari-svg#21

# TODO

- [x] ordering/prioritizing writers 
- [x] file dialog support
   - [x] file extensions/writers suggested according to layer types
   - [x] selected extension in save dialog is used to write
- [x] support napari builtins  
- [x] support binding to single-layer and multi-layer hooks from original plugin system.
- [x] tests
- [x] customize "save as" text without using short_title. Use `writer.save_dialog_title`
- [x] writers can declare their api spec

# Current design

Given a set of layers,  we need to be able to select one or more compatible writers. Ideally, these are used to present the user with a list of choices in the "Save As" file dialog, enabling the user to pick their favorite. In the case where the user doesn't indicate a preference, a predictable writer choice should be made. 

Current (before this PR) Napari behavior is to invoke installed writer plugins one-by-one until the first succeeds. A set of built-in plugins is tried last.  The intent is that these builtins provide reasonable fallbacks so all layer type combinations may be written.

### Compatibility

Currently, the pluggy-style hook specifications define two kinds of writers: single-layer writers and a multi-layer writer.

The `WriterContributions` assumes the command being bound corresponds to a multi-layer `writer()` function such as the one returned by a `napari_get_writer()` implementation. 

Alternatively, the command could correspond to one of the single-layer writers, e.g . `napari_write_image`.  This is indicated in the `WriterContribution` when at most one layer matches the layer type constraints.  For example:

```yaml
contributions:
  commands:
  - id: napari_builtins.write_points
    python_name: napari.plugins._builtins:napari_write_points
    title: napari built-in points writer
  writers:
  - command: napari_builtins.write_points
    filename_extensions: ['.csv']
    layer_types: ['points']
    save_dialog_title: napari points
```

There is no need to bind a function like `napari_get_writer()` in npe2 because writers use the manifest to declare which writer npe2 should select.  That is, npe2 has the responsibility for mapping a list of layer types to a writer.

### Layer type constraints

A writer plugin can declare that it will write between *m* and *n* layers of a specific type where *0* <= *m* <= *n*.

For example:

```
    image      Write exactly 1 image layer.
    image?     Write 0 or 1 image layers.
    image+     Write 1 or more image layers.
    image*     Write 0 or more image layers.
    image{k}   Write exactly k image layers.
    image{m,n} Write between m and n layers (inclusive range). Must have m<=n.
```

When a type is not present in the list of constraints, that corresponds to a writer that is not compatible with that type. For example, a writer declaring:

```
    layer_types=["image+", "points*"]
```

would not be selected when trying to write an `image` and a `vector` layer because the above only works for cases with 0 `vector` layers.

Note that just because a writer declares compatibility with a layer type does not mean it actually writes that type.  In the example above, the writer might accept a set of layers containing `image`s and `point`s, but the write command might just ignore the `point` layers. The writer must return `None` for unwritten layers.

### Writer selection

Most of the writer selection logic is on the napari side.

Given a list of layers of different types, the number of each type is counted and compared against the constraint ranges specified for each writer.  This yields a set of compatible writers.

These writers can be used to populate the "Save As" file dialog with a list of compatible writer commands and their file extensions. Ultimately the user selects a specific writer command to invoke via this dialog.

When a user doesn't pick a specific writer (e.g when saving from the command line), the situation is more complicated. This PR does the following:

1. When the path has a file extension, find a compatible writer that has that same extension.

2. When there is no extension and only a single layer, take the first compatible writer and append the first listed file extension.

3. Otherwise, find a compatible no-extension writer and write to that. No-extension writers typically write to a folder -- at least that's what the built-in does.


npe2/_plugin_manager.py
```diff
@@ -133,6 +153,19 @@ def iter_compatible_readers(
                         seen.add(r.command)
                         yield r
 
+    def iter_compatible_writers(
```
tlambert03, line 156: would be nice to unify this a bit with the reader pattern.  (we can make it more like this, or this more like it... but here's some diffs):

`self._readers` is a mapping of filename pattern to `ReaderContribution`, so `iter_compatible` readers can simply yield the reader itself as it finds a match.  Here `_writers_by_extension` and `_writers_by_type` store the command, and then have to reconstruct a WriterContribution instance here.  can we maybe just store the contribution itself?

npe2/manifest/io.py
```diff
@@ -15,3 +15,63 @@ class ReaderContribution(BaseModel):
...
+
+        Rules:
+        3. File extensions must start with '.' or '*.'
+        4. File extensions must be at least two characters long.
+        """
+
+        exts = [e if e[0] != "*" else e[1:] for e in exts if len(e) > 1]
+        exts = [e if e[0] == "." else f".{e}" for e in exts]
+
+        if any(True if len(e) < 1 else e[0] != "." for e in exts):
```
tlambert03, line 69:
```suggestion
        if not all(e.startswith('.') for e in exts):
```

(won't the length issue get caught below?)

npe2/_plugin_manager.py
```diff
@@ -43,6 +44,11 @@ class PluginManager:
     _themes: Dict[str, ThemeContribution] = {}
     _readers: DefaultDict[str, List[ReaderContribution]] = DefaultDict(list)
 
+    _writers_by_extension: DefaultDict[str, List[ReaderContribution]] = DefaultDict(
```
tlambert03, line 47: `ReaderContribution` -> `WriterContribution`

npe2/_plugin_manager.py
tlambert03: how would we indicate: "I don't care what extension you have to use... I just want you to save these points for me please"?

npe2/_plugin_manager.py
```diff
@@ -67,6 +75,18 @@ def discover(self):
                         self._readers[pattern].append(reader)
                     if reader.accepts_directories:
                         self._readers[""].append(reader)
+                for writer in mf.contributes.writers or []:
```
tlambert03, line 78: this would be  an annoying writer... but what about a writer that supports '.tif', but only for images, and '.csv', but only for points?  they might have filename_extensions = ['.csv', ',tif'] and layer_types = ['image', 'points'] right?  I know that ideally, that would be two separate writers, but i guess this is part of the complication with writers ... we need to have a very clear idea about what a "unit" writer plugin is?  Either we need to somehow explicitly prohibit that sort of thing, or we need to restrict the spec somehow to a single extension per group of layer types?

npe2/_plugin_manager.py
tlambert03: thinking more here, what about a writer that can write you a `.xyz` file, but _only_ if you've given them _both_ an image and a points layer... (like a STORM file or something like that).  Most likely, that would be a case where the user picks the "plugin", more than they pick the extension.  but it'd be nice to have a sense for where/how we declare that sort of thing

npe2/_version.py
```diff
@@ -1,4 +1,5 @@
+# coding: utf-8
```
tlambert03, line 1: this file should actually be added to .gitignore ... I can give this repo a general pre-commit cleanup and do that

npe2/_plugin_manager.py
tlambert03: kinda thinking out loud here... but I wonder whether this couldn't be a single dict:

```py
self._writers: DefaultDict[Tuple[FrozenSet[str], str], List] = DefaultDict(list)
for writer in mf.contributes.writers:
    layer_types = frozenset(writer.layer_types)
    for ext in writer.file_name_extensions:
        self._writers[(layer_types, ext)].append(writer)

# example
# self._writers = {
#    (frozenset({'points', 'image'}), '.xyz'): [..., ]
# }

# then later

for (types, ext), writers in self._writers.items():
    # maybe issubset... maybe == ... that's another discussion
    # and relates to my "STORM.xyz" comment below
    if set(layer_types).issubset(types) and extension == ext:
        yield from writers

```

tlambert03 (commented):
looking good!  left a lot of thoughts.  this stuff is really so hard.  If you can identify places where we should just "restrict" the spec and what it's capable of, in order to gain clarity, I'm very open to that too!

npe2/manifest/io.py
```diff
@@ -15,3 +15,63 @@ class ReaderContribution(BaseModel):
     accepts_directories: bool = Field(
         False, description="Whether this reader accepts directories"
     )
+
+
+class LayerTypes(str, Enum):
+    all = "all"
```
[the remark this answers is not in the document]
nclack: I like the None idea. I don't really like having "all" in the enum.

npe2/manifest/io.py
nclack: Ha I missed the fact that `startswith` will work for zero length strings. Thanks

npe2/manifest/schema.py
```diff
@@ -42,8 +42,7 @@ class PluginManifest(BaseModel):
     )
     # easy one... we need this.  character limit?  256 char?
     display_name: str = Field(
-        "",
-        description="The display name for the extension used in the Marketplace.",
+        "", description="The display name for the extension used in the Marketplace.",
```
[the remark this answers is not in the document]
nclack: I noticed my black seemed a little funny. I'll check.  I don't have a global config. 

The pre-commit will be super helpful.

npe2/_version.py
tlambert03: ok, I just cleaned it all up quite a bit in #4  ... can you merge/rebase at some point?

npe2/manifest/schema.py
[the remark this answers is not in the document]
tlambert03: updated, you can `pre-commit install` after merging #4

npe2/_plugin_manager.py
```diff
@@ -141,6 +174,10 @@ def __getattr__(name):
     if name == "plugin_manager":
         global _GLOBAL_PM
         if _GLOBAL_PM is None:
-            _GLOBAL_PM = PluginManager()
+            try:
+                _GLOBAL_PM = PluginManager()
+            except Exception as e:
```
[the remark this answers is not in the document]
nclack: Good question.  When I was fiddling with the PluginManifest.discover() function I had mistyped `from importlib.metadata import distributions` -- misspelled something.

That threw an import error that trickled up and handled when napari imported npe2, but that import looks like:

```python
    try:
        from npe2 import execute_command, plugin_manager
    except ImportError:
        return None
```

So, it just appeared as if npe2 wasn't working at all.  Adding the handling in this snipped helped me trace the problem.  

I felt like it was useful so I left it in.

npe2/manifest/io.py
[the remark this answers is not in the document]
nclack: (from plugin meetng) @Carreau suggested being fully explicit best here.  I agree

npe2/_plugin_manager.py
nclack: I believe this also gets solved by what I just pushed.  You setup two writers that refer to the same command:

```yaml
writers:
  - command: my_plugin.writer
    filename_extensions: ['.tif']
    layer_types: ['image+']
  - command: my_plugin.writer
    filename_extensions: ['.csv']
    layer_types: ['points+']
```

In the file dialog, I believe this would end up being two separate rows in the format selection combo-box.

npe2/_plugin_manager.py
```diff
@@ -138,6 +170,106 @@ def iter_compatible_readers(
...
+        def get_candidates(layer_type: LayerType) -> Set[WriterContribution]:
+            # WriterContributions are hashed based on their command id.
+            return set(
+                map(
+                    lambda v: v.data,  # unbox the Interval object
+                    self._writers_by_type.get(layer_type, IntervalTree())[
+                        counts[layer_type]
+                    ],
+                )
+            )
```
tlambert03, line 197:
```suggestion
        def get_candidates(type_: LayerType) -> Set[WriterContribution]:
            # WriterContributions are hashed based on their command id.
            return {v.data for v in self._writers_by_type[type_][counts[type_]]}
```

npe2/_plugin_manager.py
```diff
@@ -1,10 +1,28 @@
...
+    Union,
+)
+from warnings import warn
+
+from intervaltree import IntervalTree
 
+from ._command_registry import execute_command
 from .manifest import PluginManifest
 
 if TYPE_CHECKING:
```
tlambert03, line 28: you can throw this under `if TYPE_CHECKING` for full parametrized `IntervalTree` support:

```py
    from typing import Generic, TypeVar

    T = TypeVar("T")

    class Interval(tuple, Generic[T]):
        begin: int
        end: int
        data: T

    class IntrvlTree(IntervalTree[T]):
        def addi(self, begin: int, end: int, data: T) -> None:
            ...

        def __getitem__(self, index: Union[int, slice]) -> Set[Interval[T]]:
            ...
```

npe2/_plugin_manager.py
```diff
@@ -42,16 +60,24 @@ class PluginManager:
     _submenus: Dict[str, SubmenuContribution] = {}
     _themes: Dict[str, ThemeContribution] = {}
     _readers: DefaultDict[str, List[ReaderContribution]] = DefaultDict(list)
+    _writers_by_type: DefaultDict[
+        LayerType, IntervalTree[WriterContribution]
```
tlambert03, line 64: (if you use the `IntrvlTree` def above in TYPE_CHECKING ,then change this hint here to `IntrvlTree[WriterContribution]`)

npe2/_plugin_manager.py
```diff
@@ -138,6 +170,106 @@ def iter_compatible_readers(
...
+        """Attempt to match writers that consume all layers."""
+
+        if not layer_types:
+            return
+
+        # First count how many of each distinct type are requested. We'll use
+        # this to get candidate writers compatible with the requested count.
+        counts = Counter(layer_types)
+
+        def get_candidates(layer_type: LayerType) -> Set[WriterContribution]:
```
tlambert03, line 188: the use of interval tree here is very slick!

npe2/_plugin_manager.py
```diff
@@ -138,6 +170,106 @@ def iter_compatible_readers(
...
+            if candidates:
+                candidates &= get_candidates(lt)
+            else:
+                break
+
+        def dbg(f):
+            def g(x):
+                v = f(x)
+                print(v)
+                return v
```
tlambert03, line 211: temporary?

npe2/_plugin_manager.py
```diff
@@ -138,6 +170,106 @@ def iter_compatible_readers(
...
+                    # more "specific" writers first
+                    sum(1 for c in writer.layer_type_constraints() if not c.is_zero()),
+                    # then sort by the number of listed extensions
+                    # (empty set of extensions goes last)
+                    len(writer.filename_extensions),
+                    # finally group related extensions together
+                    writer.filename_extensions,
+                )
+            ),
+        )
```
tlambert03, line 231: I find this a bit hard to parse visually ... perhaps:
```python
        yield from sorted(candidates, key=_writer_key)


def _writer_key(writer: WriterContribution) -> Tuple[bool, int, int, List[str]]:
    # writers with no file extensions (like directory writers) go last
    no_ext = len(writer.filename_extensions) == 0
    # more "specific" writers first
    nbounds = sum(not c.is_zero() for c in writer.layer_type_constraints())
    # then sort by the number of listed extensions (empty set of extensions goes last)
    ext_len = len(writer.filename_extensions)
    # finally group related extensions together
    exts = writer.filename_extensions
    return (no_ext, nbounds, ext_len, exts)
```

npe2/manifest/io.py
```diff
@@ -15,3 +16,193 @@ class ReaderContribution(BaseModel):
...
+    layer because the above only works for cases with 0 `vector` layers.
+
+    Note that just because a writer declares compatibility with a layer
+    type does not mean it actually writes that type.  In the example
+    above, the writer might accept a set of layers containing `image`s and
+    `point`s, but the write command might just ignore the `point` layers
+    """
+
+    layer_type: LayerType
+    bounds: Tuple[int, Optional[int]] = Field(
```
tlambert03, line 70: I seems that `bounds[1]` will _always_ be an int right?  (I don't see anywhere where it ends up being `None`).  Furthermore, I get a type error if I try `IntervalTree.addi(1, None, '...')` ... so it seems like this needs to be `Tuple[int, int]`,  yeah?

tlambert03 (commented):
this is looking awesome!  still have a ways to go with the review... but just wanted to leave some thoughts. 
I think I really like the constraints pattern here!  I wonder whether it might be _slightly_ confusing for novice plugin writers... but I'm very impressed with the backend implementation, and if anyone wants us to to simplify the front-facing syntax in the manifest, i'm sure we can do that without changing much here.  very cool

tests/sample/my_plugin/napari.yaml
```diff
@@ -12,6 +12,14 @@ contributions:
     - id: my_plugin.tiff_reader
       title: My Tiff Reader
       python_name: my_plugin:get_reader
+    - id: my_plugin.my_writer
+      title: My Multi-layer Writer
+      short_title: Multi
```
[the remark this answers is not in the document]
nclack: When populating the description in the "Save As" dialog in napari, I use the short_title if available, which works pretty well imo.

An alternative might be to have a field dedicated to that purpose as a member of the WriterContribution: something like "save_dialog_descriptor" or something.  I like having clear, narrowly-scoped behavior for these fields.  The downside is the user has to provide more, potentially duplicated redundant in the spec.

npe2/manifest/io.py
```diff
@@ -15,3 +16,193 @@ class ReaderContribution(BaseModel):
...
+        1. File extensions beginning with '*' have their leading '*' removed.
+        2. File extensions lacking a leading '.' have a leading '.' added.
+
+        Rules:
+        3. File extensions must start with '.' or '*.'
+        4. File extensions must be at least two characters long.
+        """
+
+        exts = [e if e[0] != "*" else e[1:] for e in exts]
+        exts = [e if len(e) > 0 and e[0] == "." else f".{e}" for e in exts]
```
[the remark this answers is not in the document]
tlambert03: can you commit these suggestions if you like them?  makes it easier to review the "current" state of the PR

npe2/_plugin_manager.py
nclack: Seems like I couldn't just throw this under TYPE_CHECKING, but maybe I'm fooling myself.

tests/sample/my_plugin/napari.yaml
```diff
@@ -23,6 +31,17 @@ contributions:
     - command: my_plugin.tiff_reader
       filename_patterns: ["*.tif", "*.tiff"]
       accepts_directories: false
+  writers:
+    - command: my_plugin.my_writer
+      filename_extensions: ["*.tif", "*.tiff"]
+      layer_types: ["image{2,4}","tracks{1}"]
```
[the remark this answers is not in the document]
nclack: agree.  could derive `enablement` expressions from `layer_types`.

npe2/_plugin_manager.py
nclack: Oh, I'd get a "not defined" constructor at runtime.  For example, running the tests gives

```
npe2/__init__.py:9: in <module>
    from ._plugin_manager import write_layers
npe2/_plugin_manager.py:73: in <module>
    class PluginManager:
npe2/_plugin_manager.py:81: in PluginManager
    ] = DefaultDict(TypedIntervalTree)
E   NameError: name 'TypedIntervalTree' is not defined
```

tlambert03: oh also... the argument to `DefaultDict()` should _not_ be your typed version ... (just use it for the type hint).  So, something like this:

```py
from __future__ import annotations

...

_writers_by_type : DefaultDict[LayerType, TypedIntervalTree[WriterContribution]] = DefaultDict(IntervalTree)
```

npe2/manifest/io.py
```diff
@@ -15,3 +16,199 @@ class ReaderContribution(BaseModel):
...
+
+    For example:
+
+    ```
+        image      Write exactly 1 image layer.
+        image?     Write 0 or 1 image layes.
+        image+     Write 1 or more image layers.
+        image*     Write 0 or more image layers.
+        image{k}   Write exactly k image layres.
+        image{m,n} Write between m and n layers (inclusive range). Must have m<=n.
```
sofroniewn, line 48: very nice!

tests/sample/my_plugin/napari.yaml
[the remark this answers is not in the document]
nclack: I no longer use short_title and instead added a field to the writer `writer.save_dialog_title`.

sofroniewn (commented):
Looking good!!

PR #5, opened 2021-09-28
Misc validation and testing.

Carreau:


tests/test_npe2.py
```diff
@@ -51,3 +52,40 @@ def test_cli(monkeypatch):
...
+    return data
+
+
+def _mutator_4(data):
+    assert "contributes" in data
+    data["contributes"]["commands"][0]["python_name"] = "this.has.no.colon"
+    return data
+
+
+@pytest.mark.parametrize("mutator", [_mutator_1, _mutator_2, _mutator_3, _mutator_4])
```
tlambert03, line 82: ❤️

tlambert03 (commented):
good stuff!

Carreau:
marking as ready, it may conflicts with #3 so I'm happy to get 3 in first.

tlambert03 (approved):
happy to have this in now.  I'll leave it up to @nclack whether he'd prefer not having to merge or not?

npe2/manifest/commands.py
```diff
@@ -1,3 +1,4 @@
+from textwrap import dedent
```
nclack, line 1: should textwrap get added to the dependencies?

nclack (approved):
Would love to have this in now.

npe2/manifest/commands.py
Carreau: no that stdlib : https://docs.python.org/3/library/textwrap.html

nclack: yay

PR #8, opened 2021-09-29
Rename contributes to contributions ?

Carreau:
Most (all) keys are nouns, `contributes` is the only verb – I think, i'm
not a linguist.

I think for consistency it should be a noun as well.

I know that vscode uses contributes, so we can also decide to keep
contributes.

It may also be that the contributes feels weird to me as a non-native
speaker and that it's ok for a native one ? SO happy to not go this
route.

### DIFF
--- a/EXAMPLE/napari.json
+++ b/EXAMPLE/napari.json
@@ -1,7 +1,7 @@
 {
   "name": "my_plugin",
   "entry_point": "main.py",
-  "contributes": {
+  "contributions": {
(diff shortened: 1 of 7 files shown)


tlambert03:
I'm for the change too.

tlambert03 (approved):

tlambert03:
👆 (conflicts now)

Carreau:
Rebased.

PR #10, opened 2021-09-29
Rename command command field to id.

### LINKED ISSUE #9: Rename command's "command" to id.
I think that the `command` field of each command is misleading. Especially when submenu also have a `command` field
and one of the `command` field defines an id, and another one refers to an existing id.

Again I know that it's what was code does, but I believe it would be more understandable to rename the command id field to just `id`, or something else not ambiguous.

Carreau:
Closes #9


tlambert03 (approved):
like it!

PR #11, opened 2021-09-29
Better pytest error on invalid schema.

Carreau:
If a schema is invalid `.discover()` will just ignore it, which will
lead to test suite saying `len([])!=1` for example,
here I add an option to discover() to raise if it finds an invalid
manifest, which gives useful errors.

I came across that when trying to rename commands command to id


Carreau:
An alternative would be to use warning.warn and a warning filter into error, which we can do with pytest.

Any preference ?

tlambert03:
> An alternative would be to use warning.warn and a warning filter into error, which we can do with pytest.

hmm, I'm torn between the two options.  I have a slight preference for the warning, and I guess we can use `filterwarnings` elsewhere if we ever wanted to suppress it (for user-facing reasons).  I think i went with logging for that reason?  I dunno... what do you tihnk

Carreau:
I'm torn as well, I'm wondering if we want `discover()`, to return a tuple of loaded/failed to load to handle it in the UI.

Using warning.warn would also help to emit toast in the UI...

Carreau:
what does @nclack thinks ?

tlambert03:
> I'm wondering if we want discover(), to return a tuple of loaded/failed to load to handle it in the UI.

This seems like maybe the best thing to do for now.  both easy to test, and leaves more possibilities open

nclack:
I love the idea of returning a list of loaded/failed for all kinds of reasons.  I'm generally a fan of more logging/issuing warnings.

tlambert03:
ok, @Carreau ... go for the "return some repr of a failed manifest" approach.  I guess you could collect the EntryPoint objects?

Carreau:
Ok, reworked, I'm not supper happy with it. 
Discover yield either (manifest, None), or (None, exception).

npe2/manifest/schema.py
```diff
@@ -282,14 +291,16 @@ def discover(cls, entry_point_group=ENTRY_POINT) -> Iterator[PluginManifest]:
...
+                    yield pm, None
+                except ValidationError as e:
                     logger.warning(msg=f"Invalid schema {ep.value!r}")
+                    yield None, e
                 except Exception as e:
                     logger.warning(
                         "%s -> %r could not be imported: %s"
                         % (entry_point_group, ep.value, e)
                     )
+                    yield None, e
```
tlambert03, line 303: it would be nice here, and above, to associate the entry point object `ep` as well.   should it be a tuple of `(e, ep)`?

tlambert03 (approved):
fine by me.  better than what we currently have.  left a small question asking whether we should somehow remember the bad entry point.. but it's a minor detail.  will merge later today

npe2/manifest/schema.py
nclack: +1.  I wouldn't mind having a type encapsulating `(e,ep)` to make it a little easier on the type signatures.

Carreau:
Ok, I made it a 3 named tuple with type annotation.

PR #14, opened 2021-10-06
More Validation.

Carreau:
Make sure that we prevent extra fields, in particular prevent typos,
and test more failures/valid cases, in particular the usage of
contributes instead of contributions, and having an empty contributions.

### DIFF
--- a/tests/test_npe2.py
+++ b/tests/test_npe2.py
@@ -102,7 +102,7 @@ def _mutator_no_contributes_extra_field(data):
     """
     Contributions used to be called contributes.
 
-    Check that an extra filed fails.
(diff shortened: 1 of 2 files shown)


tests/test_npe2.py
```diff
@@ -79,13 +80,65 @@ def _mutator_4(data):
...
+    data["contributions"] = None
+    return data
 
-    import json
+
+def _mutator_no_contributes_extra_field(data):
+    """
+    Contributions used to be called contributes.
+
+    Check that an extra filed fails.
```
tlambert03, line 105:
```suggestion
    Check that an extra field fails.
```

PR #15, opened 2021-10-06
Prevent extra fields in Commands.

Carreau:


Carreau:
Yeah, there might be a global setting. Not sure.

tlambert03 (approved):

PR #18, opened 2021-10-19
General refactor, Exectuable mixin, io_utils APIs, remove some globals

tlambert03:
moving the reader function from napari

update: this is turning into a minor organizational refactor.  I'm trying to reduce the number of global objects, increase type hinting, and increase test coverage.


tlambert03:
@nclack, @Carreau 

apologies, this is a schmorgesborg of little refactors.  If there's anything you'd like me to pull out for futher consideration let me know.  but one thing led to another :joy:...

this...

- removes global PluginManager instantiation in `__getattr__`, in favor of `PluginManager.instance()`
- removes global `CommandRegistry` object, `PluginManager` now takes/creates a command registry.
- factors out all of the `PluginManager._<index>` dicts into a `_ContributionsIndex` object (these are things that make for easier lookup later.
- removes the global `npe2.register_command` function (which we arent' really using anyway), and puts it on the plugin context that gets passed to the `activate()` function.
- creates a `types` module for oft-used types.
- creates `io_utils.py` for `npe2.read` and `npe2.write_layers` functions, (i.e. public functions that let you use napari plugins without napari)
- changes the write_layers API a bit, (minimizes what napari needs to do downstream to use it).  cc @nclack.
- makes the `Executable` mixin so that that any contribution that points to a `command` id can be executed. for example `reader.exec()`.  This also allows for better command-specific type hinting.
- better isolation of the plugin manager from the environment during tests
- small tests fixes and more test coverage in general

tlambert03:
any thoughts on this generally @nclack?  Should i feel free to merge when ready?  or did you want to give it a close combing?  it will likely need some update on the napari side too... so if you'd prefer to have that PR ready to go as well, I can do that too

npe2/io_utils.py
```diff
@@ -0,0 +1,201 @@
...
+from typing_extensions import Literal
+
+from . import PluginManager
+from ._types import FullLayerData, LayerData, PathLike
+
+if TYPE_CHECKING:
+    from .manifest.io import ReaderContribution, WriterContribution
+
+
+def read(path: PathLike, *, plugin_name: Optional[str] = None) -> List[LayerData]:
```
nclack, line 14: I'd prefer `read_layer_data`.  Something longer is easier to search for and it seems this function and `write_layer_data` should have similar names.

tlambert03: Do we ever expect to be reading writing anything besides layer data here?  If not, feels like npe2.read and npe2.write are much nicer.  We can always add an alias though and both be happy :)

npe2/manifest/schema.py
```diff
@@ -299,34 +316,36 @@ def discover(cls, entry_point_group=ENTRY_POINT) -> Iterator[DiscoverResults]:
 
         [1]: https://packaging.python.org/specifications/entry-points/
 
+        Parameters
+        ----------
+        entry_point_group : str, optional
+            name of entry point group to discover, by default 'napari.manifest'
+        paths : Sequence[str], optional
+            sequ, by default ()
```
nclack, line 324: This line doesn't look right

npe2/_plugin_manager.py
```diff
@@ -52,125 +53,116 @@ def __getitem__(self, index: Union[int, slice]) -> Set[Interval[T]]:
...
+        if ctx._activated:
+            # prevent "reactivation"
+            return ctx
+
         try:
             modules_pre = set(sys.modules)
-            pm.activate(ctx)
+            mf.activate(ctx)
             # store the modules imported by plugin activation
             # (not sure if useful yet... but could be?)
```
nclack, line 145: just looking at some of these comments. Maybe we should get rid of ctx._imports etc if we're not using it.

nclack (approved):
Looks good!  The main thing for me here is the name of the `read` function.

npe2/io_utils.py
```diff
@@ -0,0 +1,201 @@
...
+        if plugin_name and not rdr.command.startswith(plugin_name):
+            continue
+        read_func = rdr.exec(kwargs={"path": path})
+        if read_func is not None:
+            try:
+                layer_data = read_func(path)
+                if layer_data:
+                    return (layer_data, rdr) if return_reader else layer_data
+            except Exception:
+                # TODO: what here.
```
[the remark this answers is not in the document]
tlambert03: >  if the reader returns a callable, that callable should behave

yeah, I'm ok with this too.  The current behavior is to try to go to the next one (unless a specific plugin was requested)... but that's caused more trouble than it's worth.  so let's just raise the exception and let napari deal with it

npe2/manifest/schema.py
tlambert03: lol

npe2/_plugin_manager.py
tlambert03: sure, will do

npe2/io_utils.py
nclack: Based on the types, these functions look specific to layer data.  And we don't do anything more general right now, so I lean towards the more specific longer name.

I do like short verbs. I think I'd lean more towards the short name if this were going to be used on the command line more, but usually these will get invoked inside napari.

Really, either way is fine as long as the names are self consistent: both short or both long.

tlambert03: > but usually these will get invoked inside napari.

no, that's kind of the whole point!  As we've discussed a few times, I'm specifically seeing these functions as the solution to https://github.com/napari/napari-plugin-engine/issues/24

tlambert03: > Based on the types, these functions look specific to layer data. And we don't do anything more general right now

... and I can't imagine we ever would.  It's that, combined with my desire to finally solve napari/napari-plugin-engine#24 that's making me suggest this.  (if it were just for napari to use, I wouldn't really care at all what we call it)

nclack: > I can't imagine we ever would.

oh that's interesting.  The "more" that I'm thinking of is something like the napari configuration: settings + selected ui state + layer data.  But, if I'm understanding, that would be very napari specific and not done in this plugin engine.

tlambert03: > But, if I'm understanding, that would be very napari specific and not done in this plugin engine.

oh yeah def.  napari uses npe2, npe2 doesn't use napari.  npe2 will likely never know about UI state

nclack: alright I'm sold on the short names.

PR #20, opened 2021-10-20
Manifest cli

ppwadhwa:
I have gotten a few things in place for the manifest cli.  It still needs some work, but I thought I'd go ahead and open this PR, to see if you all have some comments about where I'm grabbing some of the information below.


npe2/manifest_cli.py
```diff
@@ -0,0 +1,45 @@
...
+                publisher=publisher,
+                display_name=display_name,
+                description=description,
+                version=version,
+                contributions=contributions,
+            )
+
+            return pm.yaml()
+
+    return None
```
tlambert03, line 45:
```suggestion
from npe2.manifest import PluginManifest


def create_manifest(plugin_name: str) -> str:
    """function that takes plugin name and exports manifest"""

    from napari.plugins import plugin_manager as plugin_manager

    plugin_manager.discover()

    if plugin_name not in plugin_manager.plugins:
        raise ValueError("Could not find plugin", plugin_name)

    module = plugin_manager.plugins[plugin_name]
    standard_meta = plugin_manager.get_standard_metadata(plugin_name)
    package = standard_meta["package"].replace("-", "_")

    commands = []
    for caller in plugin_manager._plugin2hookcallers[module]:
        for impl in caller.get_hookimpls():
            if impl.plugin_name != plugin_name:
                continue
            name = impl.specname.replace("napari_", "")
            id = f"{package}.{name}"
            python_name = f"{impl.function.__module__}:{impl.function.__qualname__}"
            title = " ".join(name.split("_")).title()
            commands.append({"id": id, "python_name": python_name, "title": title})

    pm = PluginManifest(
        name=package,
        publisher=standard_meta["author"],
        description=standard_meta["summary"],
        version=standard_meta["version"],
        contributions={"commands": commands},
    )
    return pm.yaml()
```

i was just playing with this locally, sorry for the large suggestion.  here are the bullet points:

- don't import napari until inside of the function
- let's not accept a plugin manager, just always use the napari.plugins.plugin_manager
- we should call `discover()` manually to make sure that everything is found before we start
- we should raise an exception if something isn't found (rather than returning None)
- the manifest name should be the package name, _not_ the current plugin name in napari (confusing)
- I'm iterating through `plugin_manager._plugin2hookcallers[module]` rather than `plugin_manager.get_standard_metadata(plugin_name)["hooks"]` ... the main reason is that it's conceivable that a plugin has used a different name than the hookspec (i.e. you don't _have_ to name your function `napari_get_reader` ... you can actually name it something else...), so we really need to grab the exact handle to the function, which only the hookCaller can give us.

everything else looks good!  

you should start to install various plugins and then do things like 

```py
from npe2.manifest_cli import create_manifest
from napari.plugins import plugin_manager

plugin_manager.discover()

for p in plugin_manager.plugins:
    try:
        mf = create_manifest(p)
        print(p, 'worked!')
    except Exception as e:
        print(p, "failed", e)
```

ppwadhwa:
@tlambert03 @nclack I have added the changes that Nathan and I went over yesterday to try to get plugin information from the napari_plugin_engine and not napari.  I was not able to get it to work.  It is not discovering plugins. If I do it from napari, it works, but something is missing here.  I'll keep looking at it, but if you either of you see the problem.  let me know!

tlambert03:
hey @ppwadhwa, I pushed some changes directly to this PR, hope you don't mind.
The main issue I suspect you were having is that you didn't [pass discover_entry_point](https://github.com/tlambert03/npe2/pull/20/files#diff-3cd452f57cf0f01885215f041f08a433647e234f498ea8a47f7776c8f0f8b1faR32) to the `PluginManager` ... (simply naming it 'napari' isn't enough for it to know how to find plugins).

also, I removed the hookspec file and types.  For our purposes, [this class](https://github.com/tlambert03/npe2/pull/20/files#diff-3cd452f57cf0f01885215f041f08a433647e234f498ea8a47f7776c8f0f8b1faR7-R23) will suffice.

the next things that would be nice:

1. add additional things to the manifest, like writers, readers, samples, etc...
2. try to be a bit more robust to errors.  Currently, if someone tried `npe2 convert napari-svg` instead of `npe2 convert svg` it would fail (and they'd see that "svg" was in the list of detected plugins).  but we should be able to do better than that, but checking for a package by that name, and looking at its entry points for a `napari.plugin` entry.
See [here](https://github.com/tlambert03/npe2/blob/f88f1a9bbe3921f6119d949cb5bba39090b628a3/npe2/manifest/schema.py#L178-L180) for the basic idea

npe2/manifest/schema.py
```diff
@@ -376,6 +378,44 @@ def _from_package_or_name(
 
     ValidationError = ValidationError  # for convenience of access
 
+    @classmethod
```
tlambert03, line 381: put this here, but i'd be equally happy with putting it in `_from_npe1` and import PluginManifest there

npe2/manifest/schema.py
```diff
@@ -376,6 +378,84 @@ def _from_package_or_name(
...
+                py_name = f"{impl.function.__module__}:{impl.function.__qualname__}"
+                title = " ".join(name.split("_")).title()
+                commands.append({"id": id, "python_name": py_name, "title": title})
+
+                if "reader" in id:
+                    # add this to the readers list
+                    readers.append({"command": id, "accepts_directories": True})
+
+                write_layers = []
+                write_directory = False
```
tlambert03, line 417:
```py
if "write" in id and "get_writer" not in id:
```
(rather than nested if)

npe2/manifest/schema.py
```diff
@@ -376,6 +378,84 @@ def _from_package_or_name(
...
+                            writers.append(
+                                {"command": id, "layer_types": [layer], "name": layer}
+                            )
+                            write_layers.append(layer + "*")
+                        else:
+                            id_dir = id
+                            write_directory = True
+                if write_directory:
+                    if len(write_layers) > 0:
+                        writers.append({"command": id_dir, "layer_types": write_layers})
```
tlambert03, line 432: technically, `id_dir` is potentially unbound here.  can you reformulate this logic?

npe2/manifest/schema.py
```diff
@@ -376,6 +378,84 @@ def _from_package_or_name(
...
+                    # add this to the readers list
+                    readers.append({"command": id, "accepts_directories": True})
+
+                write_layers = []
+                write_directory = False
+                if "write" in id:
+                    # add this to the writers list
+                    if "get_writer" not in id:
+                        layer = id.split("write_")[1]
+                        if layer != "directory":
```
tlambert03, line 422: `napari_write_directory` is a not a hookspec, so won't `id.split("write_")[1]` != 'directory'` _always_ be true?

(meaning `if write_directory` will never happen)?

npe2/manifest/schema.py
```diff
@@ -376,6 +378,84 @@ def _from_package_or_name(
...
+                            write_directory = True
+                if write_directory:
+                    if len(write_layers) > 0:
+                        writers.append({"command": id_dir, "layer_types": write_layers})
+                    else:
+                        writers.append({"command": id_dir, "layer_types": ["*"]})
+
+                themes = []
+                if "theme" in id:
+                    from napari.utils.theme import get_theme
```
tlambert03, line 438: don't import napari here.  see the [specification for this hookspec](https://napari.org/plugins/stable/hook_specifications.html#gui-hooks).  It's a function that returns a dict of `{theme_name: theme_dict}`.  So you can just call the function directly and add each theme:
```python
for name, theme in impl.function():
    # probably safer to cast to manifest.themes.ThemeColors here to make sure the keys match
    themes.append({"label": name, id:  "type": "dark", "colors": theme})
```

npe2/manifest/schema.py
ppwadhwa: I got this from looking at this: https://github.com/napari/napari/blob/main/napari/builtins.yaml

Its not anywhere on the old napari plugin engine?  I will remove it.

nclack: the "directory writer" is basically I name I made up when authoring builtins.yaml.  The corresponding thing in the original plugin engine is the napari_get_writer implementation [here](https://github.com/napari/napari/blob/c01794a3faf11c6c8f5b9833aae55b38a24bd43b/napari/plugins/_builtins.py#L259). It happens to work by creating a directory and saving layers inside of it.  

But the question here is what to do with `napari_get_writer` hooks.  

It's not clear how to translate these automatically. The writer contribution basically binds the things that `napari_get_writer` returns rather than `napari_get_writer` itself.

npe2/manifest/schema.py
```diff
@@ -376,6 +379,87 @@ def _from_package_or_name(
...
+
+                themes = []
+                if "theme" in id:
+                    for name, theme in impl.function():
+                        # cast theme into manifest.themes.ThemeColors dict
+                        manifest_theme = ThemeColors().dict()
+                        manifest_theme = {
+                            key: theme.get(key, manifest_theme[key])
+                            for key in manifest_theme
+                        }
```
tlambert03, line 440: actually, I think you can just do `ThemeColors(**theme)` here

ppwadhwa:
@tlambert03 I made updates based on your comments above.  Both for writers and themes for the manifest, as well as a check for package names in addition to plugin names when converting.  

@nclack I see your comment above about `get_writer`.  The way I left things for now is that if there is a writer available for any layer types, then a line in the writers section will be created with the write_directory command.  This command is also added to the commands list.

I have another question about the `python_name` created.  I'm including the manifest for svg. Is  the `python_name` correct?  Also, how does the rest look?  I need to fix the list in write_directory layer types.

```
contributions:
  commands:
  - id: napari_svg.get_writer
    python_name: napari_svg.hook_implementations:napari_get_writer
    title: Get Writer
  - id: napari_svg.write_image
    python_name: napari_svg.hook_implementations:napari_write_image
    title: Write Image
  - id: napari_svg.write_labels
    python_name: napari_svg.hook_implementations:napari_write_labels
    title: Write Labels
  - id: napari_svg.write_points
    python_name: napari_svg.hook_implementations:napari_write_points
    title: Write Points
  - id: napari_svg.write_shapes
    python_name: napari_svg.hook_implementations:napari_write_shapes
    title: Write Shapes
  - id: napari_svg.write_vectors
    python_name: napari_svg.hook_implementations:napari_write_vectors
    title: Write Vectors
  - id: napari_svg.write_directory
    python_name: napari_svg.hook_implementations:write_layer_data_with_plugins
    title: Save to folder
  readers: []
  themes: []
  writers:
  - command: napari_svg.write_image
    layer_types:
    - image
    name: image
  - command: napari_svg.write_labels
    layer_types:
    - labels
    name: labels
  - command: napari_svg.write_points
    layer_types:
    - points
    name: points
  - command: napari_svg.write_shapes
    layer_types:
    - shapes
    name: shapes
  - command: napari_svg.write_vectors
    layer_types:
    - vectors
    name: vectors
  - command: napari_svg.write_directory
    layer_types:
    - image*
    - labels*
    - points*
    - shapes*
    - vectors*
description: A plugin for reading and writing svg files with napari
name: napari_svg
publisher: Nicholas Sofroniew
version: 0.1.5
```

npe2/manifest/schema.py
ppwadhwa: ok, will change this

npe2/manifest/schema.py
nclack: I prefer moving it to _from_npe1.py and keeping schema.py more specific to the npe2 schema.

tlambert03:
quick note, after conversation with @nclack:
there will be some things (like `napari_get_writer`) for which it's going to be very hard to extract the proper stuff from the old approach.  for example, there were a few things where plugins where supposed to check extensions _inside_ the body of their functions, and perhaps return a function (like from `napari_get_writer`).  In those cases, we won't be able to easily convert to the new manifest.

I suggest tracking things like that, and having a message print after running `npe2 convert` that points the user towards the fields they should manually inspect.  You could perhaps even use strings like `'<EDIT ME>'` in places in the manifest that they should look:

```yaml
  - command: napari_svg.write_labels
    layer_types:
    - labels
    name: labels
    filename_extensions: ['<EDIT ME>']
```

tlambert03:
I updated this to work for sample, dock_widget, and function hookspecs.  I refactored it into a visitor-pattern class, as it was getting a bit hairy.  added more tests

tlambert03:
I'd like to merge this soonish, since I think it's ready for many cases, and we can improve the rest later.  @nclack @ppwadhwa, could you take another look and let me know if you want anything else done before merge?  I think `napari_get_writer` is the last hookspec to deal with, but it can come in a follow up too

ppwadhwa:
I tried it this weekend-- at least to create the yaml. It worked great! I
also made a plugin with cookiecutter, and put a reader in place.  I created
the manifest of that plugin, and need to see that napari uses it
successfully.   I had not gotten that part yet, but it is probably just
me.

I will look more closely this evening at code, but from the quick look I
did already, it looked good.

npe2/_from_npe1.py
```diff
@@ -0,0 +1,327 @@
...
+        # let these raise exceptions here immediately if they don't validate
+        cmd_contrib = CommandContribution(
+            id=cmd, python_name=py_name, title=f"Create {wdg_name}"
+        )
+        wdg_contrib = WidgetContribution(command=cmd, name=wdg_name)
+        self.contributions["commands"].append(cmd_contrib)
+        self.contributions["widgets"].append(wdg_contrib)
+
+    def napari_get_writer(self, impl: HookImplementation):
+        return NotImplemented
```
nclack, line 263:
```suggestion
    def napari_get_writer(self, impl: HookImplementation):
        warnings.warn(
            "Found a multi-layer writer, but it's not convertable. Please add the writer manually."
        )
        return NotImplemented
```

tests/test_conversion.py
```diff
@@ -0,0 +1,92 @@
+from magicgui import magic_factory
+from napari_plugin_engine import napari_hook_implementation
```
nclack, line 2:
```suggestion
from importlib.metadata import PackageNotFoundError
import pytest
from magicgui import magic_factory
from napari_plugin_engine import napari_hook_implementation
```
For a test I'm suggesting below.

tests/test_conversion.py
```diff
@@ -0,0 +1,92 @@
...
+        return some_function
+
+
+def test_conversion():
+    assert manifest_from_npe1("svg")
+
+
+def test_conversion2():
+    mf = manifest_from_npe1(module=TestPlugin)
+    assert isinstance(mf.dict(), dict)
```
nclack, line 92: Suggesting a test to get a bit more coverage over the code handling what happens when the module isn't found:

```suggestion
def test_conversion2():
    mf = manifest_from_npe1(module=TestPlugin)
    assert isinstance(mf.dict(), dict)


def test_conversion_missing():
    with pytest.raises(PackageNotFoundError):
        manifest_from_npe1("does-not-exist-asdf6as987")
```

nclack (approved):
This looks great and the visitor pattern makes it easy to understand/play with.

Added a couple of suggestions. 

One of them was to use a warning to notify the user that napari_get_writer is getting skipped. What do you think about adding warnings more generally when there's something the user needs to edit?

This is working as is so I'm approving. Follow ups can go in new issues imo.

tlambert03:
> What do you think about adding warnings more generally when there's something the user needs to edit?

yep!  this is actually in there already: the cli will [run the whole things with catchwarnings](https://github.com/tlambert03/npe2/pull/20/files#diff-1cc1fb18f0cdb58d66e071b6b7fc1f64e4865d479f0d22498568c8ff5aa99959R63-R65)... and print them in color after it's done.  so your new warning will join the other ones and get printed nicely.

nclack:
> print them in color

ha. I noticed I got nice color output but never stopped to wonder why. very nice.

Issue #22, opened 2021-11-03
validation for display name


nclack:
This issue proposes a task for adding a validator for `display_name`.

One proposal:
64 character, ascii only, cannot start with space, no tabs

`display_name` is presented in a few different contexts, one of which is the save file dialog.

It appears like the longest plugin name to date is 39 characters long.

nclack:
maybe a 40 character limit is better.

neuromusic:
for reference, our longest plugin name is https://github.com/haesleinhuepf/napari-segment-blobs-and-things-with-membranes & its most likely display name would be "Segment Blobs and Things with Membranes" which clocks in at 39 characters

"Allen Cell & Structure Segmenter" hits 32

neuromusic:
even if we set a standard that the napari viewer and the napari hub conform to, both will probably need to do some "cleaning" since we can't guarantee that devs won't go past the stick emoji in there or what have you.

so at best we're promising "follow these naming guidelines and we will accurately name your plugin, otherwise we will coerce tabs to spaces and drop unsupported characters and truncate your plugin's display name"

neuromusic:
comment from @liaprins-czi:

> I think 40 is good! too much beyond that and it is becoming a description instead of a name

nclack:
>  can't guarantee that devs won't go past the length limits or stick emoji in there

With the resolution of this issue, such a manifest would fail to validate and therefore wouldn't load into napari - a warning is issued if a bad manifest is encountered during discovery.

I was hoping to stay out of the cleanup business with this one.

neuromusic:
@nclack ahh, gotcha. I guess we'll still have that risk on the napari hub (since a dev could plausibly release a plugin w/ a bad manifest), but maybe in that case we use the same manifest validation and only support manifest-dependent features if it is a valid manifest

PR #23, opened 2021-11-09
Add display_name validation

nclack:
This adds validation for the display_name.

Addresses #22.

This uses a regex to limit display_name to 3-40 characters. The name beginning and end must not be whitespace or and underscore. The other characters must be "word" characters.

Play with the regex [here](https://regex101.com/r/QyanPt/1)

### DIFF
--- a/npe2/manifest/schema.py
+++ b/npe2/manifest/schema.py
@@ -71,7 +71,7 @@ class PluginManifest(BaseModel):
 
     display_name: str = Field(
         "",
-        description="The display name for the extension used in the Marketplace.",
+        description="User-facing text to display as the name of this plugin",
         # Must be 3-40 characters long, containing printable word characters,
         # and must not begin or end with an underscore, white space, or
         # non-word character.
@@ -101,19 +101,12 @@ class PluginManifest(BaseModel):
     license: Optional[SPDX] = None
 
     contributions: Optional[ContributionPoints]
-    # # this would be there only for the hub.  which is not immediately planning
-    # # to support open ended keywords
-    # keywords: List[str] = Field(
-    #     default_factory=list,
-    #     description="An array of keywords to make it easier to find the "
-    #     "extension. These are included with other extension Tags on the "
-    #     "Marketplace. This list is currently limited to 5 keywords",
-    # )
-    # the hub *is* planning on supporting categories
+
     categories: List[str] = Field(
         default_factory=list,
         description="specifically defined classifiers",
     )
+
     # in the absense of input. should be inferred from version (require using rc ...)
     # or use `classifiers = Status`
     preview: Optional[bool] = Field(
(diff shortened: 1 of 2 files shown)


npe2/manifest/schema.py
```diff
@@ -68,10 +68,14 @@ class PluginManifest(BaseModel):
         None,
         description="The publisher name - can be an individual or an organization",
     )
-    # easy one... we need this.  character limit?  256 char?
+
     display_name: str = Field(
         "",
         description="The display name for the extension used in the Marketplace.",
```
tlambert03, line 74: lets' change the word marketplace while we're at it

tests/test_npe2.py
```diff
@@ -164,6 +164,53 @@ def test_valid_mutations(mutator, uses_sample_plugin):
     PluginManifest(**data)
 
 
+@pytest.mark.parametrize(
+    "display_name",
+    [
+        "Here there everywhere and more with giggles and friends",
```
tlambert03, line 170: 😂

tests/test_npe2.py
```diff
@@ -164,6 +164,53 @@ def test_valid_mutations(mutator, uses_sample_plugin):
...
+        "_abc",
+        "abc_",
+        "abc♱",
+    ],
+)
+def test_invalid_display_names(display_name, uses_sample_plugin):
+    pm = list(PluginManifest.discover())[0]
+    assert pm.manifest
+    data = json.loads(pm.manifest.json(exclude_unset=True))
+    data["display_name"] = display_name
```
tlambert03, line 183: this feels like a lot of trouble and mocking to go through just to test this validator...  how about:

```py
field = PluginManifest.__fields__['display_name']
value, err = field.validate(display_name, {}, loc=0)
assert err is not None
```

tests/test_npe2.py
```diff
@@ -164,6 +164,53 @@ def test_valid_mutations(mutator, uses_sample_plugin):
...
+        "abc",
+        "abc䜁䜂",
+    ],
+)
+def test_valid_display_names(display_name, uses_sample_plugin):
+    pm = list(PluginManifest.discover())[0]
+    assert pm.manifest
+    data = json.loads(pm.manifest.json(exclude_unset=True))
+    data["display_name"] = display_name
+    PluginManifest(**data)
```
tlambert03, line 203: same as above, but with `assert err is None`

tests/test_npe2.py
```diff
@@ -164,6 +164,53 @@ def test_valid_mutations(mutator, uses_sample_plugin):
...
+
+
+def test_display_name_default_is_valid(uses_sample_plugin):
```
tlambert03, line 206: ?
how about just `PluginManifest(name='', entry_point='')`

npe2/manifest/schema.py
nclack: ha you got it.

tlambert03 (approved):

PR #25, opened 2021-11-09
Small updates for napari

tlambert03:
mostly just adds the `get_manifest` command back (it expects a `manifest.name`), and adds an `iter_themes` convenience.  Also adds a `plugin_name` property convenience on the Executable mixin.  And removes the manifest `key` property for now (this was the combination of publisher_name and manifest name ... but I think it's more confusing and overdetermined at the moment (since pypi should already provide uniqueness)


nclack (approved):
lgtm

PR #26, opened 2021-11-10
support for Dock Widgets

tlambert03:


npe2/manifest/widgets.py
```diff
@@ -0,0 +1,14 @@
...
+
+from .._types import WidgetCallable
+from .utils import Executable
+
+
+class WidgetContribution(BaseModel, Executable[WidgetCallable]):
+    command: str = Field(
+        ..., description="Identifier of a command that returns a Widget instance."
+    )
+    name: str
```
nclack, line 11: tests are failing because name is required but the name isn't provided in ` tests/sample/my_plugin/napari.yaml`.  Do we want to require name?

nclack: this should probably be optional

I like the pattern we ended up with on writers where `name` is optional and plugin `display_name` is used for presentation.

The logic for the save file dialog ended up looking like this (in napari):

```python
            name = pm.get_manifest(writer.command).display_name
            title = f"{name} {writer.name}" if writer.name else name
```

nclack: looking at the how `_add_plugin_actions` works on the napari side, it seems like we should add a validator here.

If there's just one widget, it's fine for it to have an empty string as a name, afaict. But if there's more than one than non-empty names are required.

nclack (approved):

PR #31, opened 2021-11-17
Sample data

tlambert03:


npe2/manifest/sample_data.py
```diff
@@ -0,0 +1,49 @@
...
+from .._types import LayerData
+from .utils import Executable
+
+if TYPE_CHECKING:
+    from .._command_registry import CommandRegistry
+
+
+class _SampleDataContribution(BaseModel, ABC):
+    display_name: str
+    key: str  # python identifier
```
nclack, line 16: maybe call this `python_name` instead?  Could also add regex validation similar to what was done for command.python_name.

nclack (approved):
looks good to me!  left one comment about a field name.

npe2/manifest/sample_data.py
tlambert03: this one isn't actually a path to a python callable (like other `python_names` are) ... it's a key used to identify this sample programmatically ... for instance `viewer.open_sample('plugin_name', 'sample_key')` ... we don't want that to be the display name, since caps and spaces and parens and all that jazz are annoying

tlambert03: added a description though to make that clearer

PR #34, opened 2021-11-21
add `get_callable` to Executable mixin

tlambert03:
the Exectuable mixin provided `exec`, which lets you call a callable based on a command id, but it didn't give an easy way to get the callable object itself (and so I missed the fact that i was returning the wrong thing for dock widgets).  this adds a `get_callable` method on the Executable mixin that returns the callable object itself


tlambert03:
this will be needed for https://github.com/napari/napari/pull/3666

nclack (approved):
looks good!

Issue #37, opened 2021-11-23
Clarify valid manifest `name` keys


DragaDoncila:
* npe2 version: 0.1.dev116+gdd529bf
* Python version: Python 3.9.7
* Operating System: MacOS Big Sur 11.6

### Description

There are some inconsistencies in the way the `name` key of the manifest is defined/parsed by `npe2`.

The docs currently mention that it should be a valid package name (or at least that is the intention as I understood from @nclack). However, based on the command validation regex, it looks like they should be module names instead? 

e.g. this manifest snippet is validated as ok by `npe2 validate`

```yaml
name: nag-demo
display_name: Workshop Demo
license: BSD-3-Clause
entry_point: nag_demo
```

but if I add a reader command referencing this name:

```yaml
name: nag-demo
display_name: Workshop Demo
license: BSD-3-Clause
entry_point: nag_demo

contributions:
  commands:
  - id: nag-demo.get_reader
    python_name: nag_demo._reader:napari_get_reader
    title: NAG Reader
  readers:
  - command: nag-demo.get_reader
    accepts_directories: false  
    filename_patterns: ['*.xfc']
```

then `npe2 validate` gives me:

```zsh
% npe2 validate src/nag_demo/napari.yaml
🅇 Invalid! 1 validation error for PluginManifest
contributions -> commands -> 0 -> id
  string does not match regex "^((([a-zA-Z_][a-zA-Z_0-9]+)\.)*([a-zA-Z_][a-zA-Z_0-9]+))$" (type=value_error.str.regex; pattern=^((([a-zA-Z_][a-zA-Z_0-9]+)\.)*([a-zA-Z_][a-zA-Z_0-9]+))$)
```

I can change the name (and associated commands) `nag-demo` to `nag_demo` (the module name) and this works fine. However, I can also change the name to any random word I want and that also still works e.g.

```yaml
name: foobar
display_name: Workshop Demo
license: BSD-3-Clause
entry_point: nag_demo

contributions:
  commands:
  - id: foobar.get_reader
    python_name: nag_demo._reader:napari_get_reader
    title: NAG Reader
  readers:
  - command: foobar.get_reader
    accepts_directories: false  
    filename_patterns: ['*.xfc']
```

gives us

```zsh
% npe2 validate src/nag_demo/napari.yaml
✔ Manifest for 'Workshop Demo' valid!
```

### Questions/issue

- the validator tool should error on a `name` key that can't be used as part of a command
[...]

Issue #29, opened 2021-11-16
Add additional validation to npe2 cli


nclack:
### Description

npe2 has a command line tool that will validate a plugin author has a well-defined manifest.

Additional validation for plugin developers would help them double check they've fulfilled functional requirements for their plugin.

This issue proposes to add features to

* do smoke testing on the callables where that's possible
* check type annotations on callables when provided
 
A flag, something like `npe2 validate --full`, should trigger these more expensive checks.

For reference, see comments at the end of napari/napari#3617.

PR #38, opened 2021-11-25
Cleanup manifest

### LINKED ISSUE #17: Remove some manifest fields
### Description

I wanted to track fields that will probably go unused or may be redundant.

These include:

- categories
- command.short_title (see https://github.com/tlambert03/npe2/pull/3#discussion_r726509561)

Presentation only:

- license
- preview

Putting these up for discussion. Please add things I missed.  Some of these may just need a use case.

nclack:
closes #17 

This removes several fields.  These include 
* contributions
* ~license~ (keep)
* categories
* preview
* private

I also went through and tried to remove comments, TODO's, add field documentation, etc.

The EXAMPLE folder was removed; it contained an old example.

Please note TODO's and FIXME's that you'd like to keep.


npe2/_command_registry.py
```diff
@@ -27,9 +27,6 @@ def register(self, id: str, command: Callable) -> PDisposable:
         if not callable(command):
             raise TypeError(f"Cannot register non-callable command: {command}")
 
-        # TODO: validate argumemnts and type constraints
-        # possibly wrap command in a type validator?
```
nclack, line 31: I'm in favor of handling this w #29 right now; the cli tool can do some initial validation to type check.

npe2/_plugin_manager.py
```diff
@@ -318,8 +317,10 @@ def get_writer(
             elif not ext and len(layer_types) == 1:  # No extension, single layer.
                 ext = next(iter(writer.filename_extensions), "")
                 return writer, path + ext
-            else:
-                raise ValueError
+            # When the list of extensions for the writer doesn't match the
+            # extension in the filename, keep searching.
+
+        # Nothing got found
```
nclack, line 323: this fixes some broken tests I was getting on the napari side. napari's CI misses these at the moment because npe2 isn't installed when the test suite is running.

npe2/manifest/schema.py
```diff
@@ -53,20 +48,17 @@ class DiscoverResults(NamedTuple):
...
 
-    # VS Code uses <publisher>.<name> as a unique ID for the extension
-    # should this just be the package name ... not the module name? (probably yes)
-    # do we normalize this? (i.e. underscores / dashes ?)
-    # TODO: enforce that this matches the package name
     name: str = Field(
         ...,
-        description="The name of the plugin - should be all lowercase with no spaces.",
+        description="The name of the plugin. Should correspond to the python "
+        "package name for this plugin.",
```
nclack, line 54: This is related to the discussion in #37. I think the best approach is to require the name to be the package name. I haven't changed any of the validation so #37 is still an issue.

tlambert03:
hmm, i didn't particularly mind most of the comments that you removed here.  I think many of them are still "open questions" that haven't received any particularly special attention that warrants comment removal. or are perhaps just generally useful in terms of where the code was inspired (such as the vscode comments).  
Can you perhaps return the comments and just focus on the fields that you want to modify?

nclack:
> Can you perhaps return the comments and just focus on the fields that you want to modify?

Yes.  Let me know if I missed anything.

I'm keeping this in sync w #40

tlambert03:
why are you removing license?

tlambert03 (approved):
outside of license, this is fine with me.  Can we leave that discussion for another PR?

I think we generally need to consider the process of taking data from setup.cfg a little better.  Let's keep it independent of "cleanup"

nclack:
> license 

added back.

> need to consider the process of taking data from setup.cfg a little better

agreed. I also like the idea of separating that discussion from this issue.

tlambert03 (approved):

PR #40, opened 2021-11-25
Change 'publisher' to 'author' (#39)

### LINKED ISSUE #39: rename 'publisher' field to 'author'
Originally, I think we we're going to use this to help distinguish plugins that had similar names.  But if we end up requiring plugin name is a python package name (see #37) then that's not necessary.

We've ended up using this more as a list of authors. When missing it's populating this from the Author field of the package metadata when it's missing.

nclack:
Closes #39

### DIFF
--- a/npe2/_from_npe1.py
+++ b/npe2/_from_npe1.py
@@ -102,7 +102,7 @@ def manifest_from_npe1(
 
     return PluginManifest(
         name=package,
-        authors=standard_meta.get("author"),
+        author=standard_meta.get("author"),
         description=standard_meta.get("summary"),
         version=standard_meta.get("version"),
         contributions=dict(parser.contributions),
--- a/npe2/_from_npe1.py
+++ b/npe2/_from_npe1.py
@@ -102,7 +102,7 @@ def manifest_from_npe1(
 
     return PluginManifest(
         name=package,
-        publisher=standard_meta.get("author"),
+        authors=standard_meta.get("author"),
         description=standard_meta.get("summary"),
         version=standard_meta.get("version"),
         contributions=dict(parser.contributions),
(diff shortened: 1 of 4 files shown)


npe2/_plugin_manager.py
```diff
@@ -52,7 +52,7 @@ def __getitem__(self, index: Union[int, slice]) -> Set[Interval[T]]:
             ...
 
 
-PluginName = str  # this is defined on PluginManifest as `publisher.name`
+PluginName = str  # this is defined on PluginManifest as `manifest.name`
```
tlambert03, line 55:
```suggestion
# this is `PluginManifest.name`
```

npe2/_from_npe1.py
```diff
@@ -102,7 +102,7 @@ def manifest_from_npe1(
 
     return PluginManifest(
         name=package,
-        publisher=standard_meta.get("author"),
+        authors=standard_meta.get("author"),
```
tlambert03, line 105:
```suggestion
        author=standard_meta.get("author"),
```

since we're doing this to match core metadata (https://packaging.python.org/specifications/core-metadata/) ... let's actually match the core metadata

npe2/manifest/schema.py
```diff
@@ -61,12 +61,11 @@ class PluginManifest(BaseModel):
         ...,
         description="The name of the plugin - should be all lowercase with no spaces.",
     )
-    # this is not something that has an equivalent on PyPI ...
-    # it might be a good field with which we can identify trusted source
-    # but... it's not entire clear how that "trust" gets validated at the moment
-    publisher: Optional[str] = Field(
+
+    authors: Optional[str] = Field(
```
tlambert03, line 65:
```suggestion
    author: Optional[str] = Field(
```

PR #44, opened 2021-12-05
PackageMetadata field

tlambert03:
this adds a PackageMetadata model, basically just a nicer object than the `email.message.Message` that comes from `importlib.metadata.Distribution`.

Adhere's strictly to PEP 566 and https://packaging.python.org/specifications/core-metadata/

### DIFF
--- a/npe2/manifest/schema.py
+++ b/npe2/manifest/schema.py
@@ -6,7 +6,6 @@ import types
 from contextlib import contextmanager
 from enum import Enum
 from importlib import import_module, util
-from importlib.metadata import Distribution
 from logging import getLogger
 from pathlib import Path
 from textwrap import dedent
@@ -29,9 +28,9 @@ from .contributions import ContributionPoints
 from .package_metadata import PackageMetadata
 
 try:
-    from importlib.metadata import distributions
+    from importlib.metadata import Distribution, distributions
 except ImportError:
-    from importlib_metadata import distributions  # type: ignore
+    from importlib_metadata import Distribution, distributions  # type: ignore
 
 if TYPE_CHECKING:
     from importlib.metadata import EntryPoint
--- a/npe2/manifest/schema.py
+++ b/npe2/manifest/schema.py
@@ -63,12 +63,6 @@ class PluginManifest(BaseModel):
         "package name for this plugin.",
     )
 
-    author: Optional[str] = Field(
-        None,
-        description="The author name(s). When unspecified, the description is "
-        "take from the 'Author' field of the package metadata.",
-    )
-
     display_name: str = Field(
         "",
         description="User-facing text to display as the name of this plugin",
@@ -98,21 +92,19 @@ class PluginManifest(BaseModel):
 
     @property
     def license(self) -> Optional[str]:
-        if self.package_metadata:
-            return self.package_metadata.license
-        return None
+        return self.package_metadata.license if self.package_metadata else None
 
     @property
     def version(self) -> Optional[str]:
-        if self.package_metadata:
-            return self.package_metadata.version
-        return None
+        return self.package_metadata.version if self.package_metadata else None
 
     @property
     def description(self) -> Optional[str]:
-        if self.package_metadata:
-            return self.package_metadata.summary
-        return None
+        return self.package_metadata.summary if self.package_metadata else None
+
+    @property
+    def author(self) -> Optional[str]:
+        return self.package_metadata.author if self.package_metadata else None
 
     @root_validator
     def _validate_root(cls, values: dict) -> dict:
@@ -335,8 +327,8 @@ class PluginManifest(BaseModel):
                     meta = PackageMetadata.from_dist_metadata(distribution.metadata)
                     mf.package_metadata = meta
 
-                    # populate missing metadata
                     mf.name = mf.name or meta.name  # should we assert?
+                    return mf
 
         raise FileNotFoundError(f"Could not find file {fname!r} in module {module!r}")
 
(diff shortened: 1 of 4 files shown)


npe2/manifest/schema.py
```diff
@@ -62,12 +62,6 @@ class PluginManifest(BaseModel):
         "package name for this plugin.",
     )
 
-    author: Optional[str] = Field(
```
tlambert03, line 65: @nclack, please note/comment on the removal of these fields.  With the package_metadata object, they seem redundant.  It's nice to remove them (to single source them), but can think of two potential issues:

- makes it a bit harder to create a manifest not based on a package (i.e. for dynamic creation).  but it's still not immediately clear to me where that will be done.  and we can always add a convenience mocker when we need it.
- makes it a bit harder for the _hub_ to parse this information statically ... since now they need to look in multiple places

npe2/manifest/schema.py
```diff
@@ -77,11 +71,6 @@ class PluginManifest(BaseModel):
         regex=r"^[^\W_][\w -~]{1,38}[^\W_]$",
     )
 
-    description: Optional[str] = Field(
-        description="A short description of what your extension is and does."
-        "When unspecified, the description is taken from package metadata."
-    )
```
nclack, line 83: This is the only one that I'm not sure about.  The package description can be a few paragraphs.  Maybe the user wants a shorter/different description for their plugin (e.g. something like napari's builtins). 

We don't actually use the description field anywhere though, so I favor removal.

npe2/manifest/schema.py
```diff
@@ -334,9 +319,16 @@ def _from_entrypoint(cls, entry_point: EntryPoint) -> PluginManifest:
...
-            if mf.exists():
-                return PluginManifest.from_file(mf)
+            mf_file = Path(loc) / fname
+            if mf_file.exists():
+                mf = PluginManifest.from_file(mf_file)
+                if distribution is not None:
+                    meta = PackageMetadata.from_dist_metadata(distribution.metadata)
+                    mf.package_metadata = meta
+
+                    mf.name = mf.name or meta.name  # should we assert?
```
nclack, line 329: assert on what exactly?  

If you're referring to `mf.name==meta.name`, I think we have to think about what to do for the napari builtins case.  I think it's ok not to assert with the caveat that users could run into namespacing issues; future conflicts w pypi packages.

npe2/manifest/schema.py
```diff
@@ -167,9 +164,7 @@ def from_distribution(cls, name: str) -> PluginManifest:
         dist = distribution(name)  # may raise PackageNotFoundError
         for ep in dist.entry_points:
             if ep.group == ENTRY_POINT:
-                pm = PluginManifest._from_entrypoint(ep)
-                pm._populate_missing_meta(dist.metadata)
-                return pm
+                return PluginManifest._from_entrypoint(ep, dist)
```
nclack, line 167: codecov is saying this didn't get hit by tests, but it does get hit on my machine

npe2/manifest/schema.py
nclack: I think the only one I want to consider keeping is "description". I made a separate comment around that.

> not immediately clear to me where that will be done. and we can always add a convenience mocker when we need it

Totally agree

> harder for the hub to parse [..] statically

I think doing this (relying on pypi) helps the hub maintain clean sources of truth for their metadata.  I'm a big fan of keeping the plugin metadata as bare bones as we can.

nclack (commented):
Left some comments, but this looks good. Thanks for putting it together!

You modelled the package metadata in detail. Is that necessary or could we restrict to a subset of "interesting" fields?

tlambert03:
> You modelled the package metadata in detail. Is that necessary or could we restrict to a subset of "interesting" fields?

Could do, but then we'd need to either "allow extra" on the model, or do do the checking ourself in `from_importlib_dist` to prune extra fields.  

I'm open to it, but curious why not just model the PEP?

npe2/manifest/schema.py
tlambert03: This is replaced by package "summary", which is _not_ the one that is many paragraphs.  See the property below on line 101.

(This was essentially another example of us using a word that didn't match setup cfg)

npe2/manifest/schema.py
tlambert03: Well, this is our chance to assert that the manifest name is the same as the package metadata.  It's a critical point obviously.  So we should either just overwrite it (using package meta), assert equality, or something?

The napari manifest is already changed to "napari", as we discussed

nclack: > The napari manifest is already changed to "napari", as we discussed

I'd forgotten. It's nice to see actually. I think it works better that way.

Now I prefer adding the assertion that they're the same.  This will avoid all kinds of confusion from typos, makes it easier to track down the plugin for commands, etc... It's better to start more strict.

nclack:
> why not just model the PEP

It's just a question of what maintenance is required when the core metadata spec changes.  Ideally there's none. It's a lot easier to add fields later than remove them. That kind of thing.  

In any case, the core metadata changes infrequently and it's a small amount of work to update so I'm not actually worried about this. It's also valuable to have easy access to the data! I was mostly wondering if there was some mechanical thing I was missing and you answered that.

It sounds like `extra=ignore` might be a good idea.

npe2/manifest/schema.py
tlambert03: asserted

nclack (approved):
looks good!

PR #46, opened 2021-12-08
Update error messages

### LINKED ISSUE #45: more user friendly error messages for regex failures
### Description

Fields that use a regex to validate have difficult to interpret error messages. For example:

```
🅇 Invalid! 1 validation error for PluginManifest
contributions -> commands -> 0 -> id
  string does not match regex "^((([a-zA-Z_][a-zA-Z_0-9]+)\.)*([a-zA-Z_][a-zA-Z_0-9]+))$" (type=value_error.str.regex; pattern=^((([a-zA-Z_][a-zA-Z_0-9]+)\.)*([a-zA-Z_][a-zA-Z_0-9]+))$)
```

It would be nice to replace these with something more friendly.

ppwadhwa:
I'm trying to improve some of the error messages that occur after failing validation in some of the manifest keys.

So far, I have included the following: 
1. A message when the `python_name` for a command doesn't match the normalized package name at the beginning of the python name. 
2. A validation of a regex for the `python_name` (to check that the whole thing is following what is expect.
3.  The validation and error message for the display name.  I didn't change the regex for this field, I just used it in a validator in order to customize the error message.
4. There is already a check in place that the command ID has the same package name at the beginning of the ID, however, I think it needs a little bit of a change.  Before, if you were to add on characters after the package name, but before the `.`, it would still pass. I made it so that the package name has to match with no trailing characters possible before the `.`

I still need to include tests.

Fixes #45


npe2/manifest/schema.py
```diff
@@ -147,14 +147,47 @@ def _check_engine_version(cls, values: dict) -> dict:
...
+
+                    if (python_name_start_expected != "dynamic") and (
+                        python_name_start_expected != python_name_start_actual
+                    ):
+                        invalid_python_names.append(command.python_name)
+
+                except (NameError, AttributeError):
+                    # There isn't a python_name here.
+                    pass
+
```
nclack, line 190:
```suggestion
```

I don't think we want this.  As far as I know, the python_name might not have anything to do with the package/plugin name.

I could be wrong! so let me know what you think.

npe2/manifest/schema.py
```diff
@@ -147,14 +147,47 @@ def _check_engine_version(cls, values: dict) -> dict:
...
     @root_validator
     def _validate_root(cls, values: dict) -> dict:
         invalid_commands = []
+        invalid_python_names = []
         if values.get("contributions") is not None:
             for command in values["contributions"].commands or []:
-                if not command.id.startswith(values["name"]):
+                id_start_actual = command.id.split(".")[0]
+                if values["name"] != id_start_actual:
```
nclack, line 170:
```suggestion
                if not command.id.startswith(values["name"]):
```

Given what I suggested below, I like the way the predicate was originally done.

nclack (changes requested):
I suggested a change to the root validator.  I could be wrong, but I don't think we need to enforce the plugin name should have to be the same as the python module name used for the callable.

npe2/manifest/schema.py
ppwadhwa: The reason I had changed this was because leaving it with `startswith` would allow something with the name `napari-svg.get_writer` to allow something `napari-svg123.get_writer`.  If we don't care that something can trail before the `.`, then this will work great.

nclack: good point! i take it back.

npe2/manifest/schema.py
```diff
@@ -147,12 +147,27 @@ def _check_engine_version(cls, values: dict) -> dict:
...
+                "be 3-40 characters long, containing printable word characters, "
+                "and must not begin or end with an underscore, white space, or "
+                "non-word character."
+            )
+        return v
+
     @root_validator
     def _validate_root(cls, values: dict) -> dict:
         invalid_commands = []
+        invalid_python_names = []
```
nclack, line 166:
```suggestion
```

npe2/manifest/schema.py
```diff
@@ -165,6 +180,17 @@ def _validate_root(cls, values: dict) -> dict:
...
+            raise ValueError(
+                dedent(
+                    f"Python names must include a normalized package name of the form "
+                    "`{obj.__module__}:{obj.__qualname__} `(e.g. "
+                    "`my_package.a_module:some_function`).  The following python names "
+                    "did not match the package names: "
+                    f"{invalid_python_names}"
+                )
+            )
+
```
nclack, line 193:
```suggestion
```

npe2/manifest/schema.py
nclack: i don't think we need this anymore. also suggested removing a related block below.

nclack (changes requested):
just one last set of suggestions

tlambert03 (approved):
LGTM!  thanks again @ppwadhwa 

i made a very minor change:

instead of 

```py
        regex = f"^{_dotted_name}:{_dotted_name}$"
        if v is not None and not bool(re.match(regex, v)):
```

since compiling the regex is the slowest part, I compiled the regex once on module import, then used:

```py
_python_name_pattern = re.compile(f"^{_dotted_name}:{_dotted_name}$")

...
        if v and not _python_name_pattern.match(v):
```

PR #49, opened 2021-12-08
Schema review

### LINKED ISSUE #48: incorporate tasks identified in schema review
### Description

Task: Incorporate changes from a review of the schema.

The review aimed to identify unused fields and other cleanup items. These aren't listed here, but will be identified as todo items added in the pr associated with this issue.

One of the cleanup items is the regex around command id's which should change to be like a python package name (rather than like a python identifier).

This task requires:
- [x] updating npe2
- [ ] review of related docs to incorporate changes
- [ ] review of the npe2 cookiecutter to incorporate changes
- [x] update the npe2 convert utility to incorporate changes (in particular around command id)

nclack:
Closes #48.  

See the issue for description and subtasks.


tlambert03:
looking good so far.  lemme know when you want a full review

nclack:
@tlambert03 I added a field for the version. Just to be difficult I called it "engine". I'm interested on what you think about that. The name takes inspiration from the `engines` field of the [vscode extention manifest](https://code.visualstudio.com/api/references/extension-manifest).

I still need to add tests around that field, but otherwise this is ready for a more detailed review.

I probably forgot some stuff. I am not renaming or otherwise changing `entry_point` right now. That should probably be a separate issue.

tlambert03:
one other thing I forgot to mention that I thought of: another way to indicate plugin manifest version prior to any parsing is to change the entry point name, e.g. from `napari.manifest` to `napari.manifest10`, etc...

npe2/_from_npe1.py
```diff
@@ -23,6 +23,7 @@
 
 from npe2.manifest import PluginManifest
 from npe2.manifest.commands import CommandContribution
+from npe2.manifest.schema import ENGINE_VERSION
```
tlambert03, line 26: why do we need to import and use this here if it's also the default value for engine?

npe2/manifest/configuration._py
```diff
@@ -0,0 +1,262 @@
+# """
```
tlambert03, line 1: what's the story here?  and what's `._py`?

npe2/_from_npe1.py
nclack: We want to have a default behavior for read, but when we're writing I think it's good to be opinionated: label the converted manifest with the version used to covert it.

npe2/manifest/configuration._py
nclack: the coverage tool was complaining and I wasn't sure if we wanted to delete the files, keep them around or what. looking for recommendations.

npe2/_from_npe1.py
tlambert03: ahh... and when we serialize it excludes not-set right?

npe2/manifest/configuration._py
tlambert03: yeah, just leave this file untouched, comment it out in Contributions, and if coverage is the only thing that fails, it's fine with me (it will only happen once)

npe2/_from_npe1.py
nclack: I guess the other part of the answer is I imagine we may not always have a default value for 'engine' or the default behavior may change. I'm guessing we always want to write the engine version though.

tlambert03: I totally agree we always want to write the engine, but i was confused why it goes in the constructor here (since it will result in an identical PluginManifest even if you excluded it).

Just so you know, since this is more of a serialization issue than an instantiation issue, another option here would be to use the `include=['engine']` parameter when serializing with `PluginManifest.json()`

npe2/manifest/schema.py
```diff
@@ -42,6 +43,12 @@
 
 ENTRY_POINT = "napari.manifest"
 
+# The first release of npe2 defines the first engine version: 0.0.0.
+# As the contract around plugins evolve the ENGINE_NUMBER should be
+# increased follow SemVer rules. Note that sometimes the version number
+# will change even though no npe2 code changes.
+ENGINE_VERSION = "0.0.0"
```
tlambert03, line 50: thoughts on `0.0.0` instead of `1.0.0` or `0.1.0` or `0.0.1` ?

npe2/manifest/schema.py
```diff
@@ -105,6 +135,20 @@ def description(self) -> Optional[str]:
     def author(self) -> Optional[str]:
         return self.package_metadata.author if self.package_metadata else None
 
+    @root_validator
+    def _check_engine_version(cls, values: dict) -> dict:
+        declared_version = semver.VersionInfo.parse(values.get("engine", ""))
+        current_version = semver.VersionInfo.parse(ENGINE_VERSION)
```
tlambert03, line 141: would prefer not to bring on a dependency just for this.  is it possible?

tlambert03 (approved):

npe2/manifest/configuration._py
nclack: reverted that commit

npe2/manifest/schema.py
tlambert03: (it's our own version string... so it's not like we need to be ready for anything beyond `version.split('.')` ... if someone puts some ridiculous unrecognized engine version in their manifest that fails to parse, i think that's a reasonable exclusion criterion)

npe2/_from_npe1.py
nclack: ah ok. I could go either way. I'm probably overthinking it. I'll leave it as is and we can revisit if it comes up

npe2/manifest/schema.py
nclack: I agree this is a bad dependency (semver's about to release 3.0 and it'll break things).

It gives us parsing and comparison which are easy to do custom.

I'm a little busy so I'd like to just make an issue to replace the dependency later so we can go ahead and merge this. But it's not a lot of work..

npe2/manifest/schema.py
nclack: I go back and forth. I like 0.1.0 too.

tlambert03:
You can merge when you want

npe2/manifest/schema.py
nclack:
```suggestion
ENGINE_VERSION = "0.1.0"
```

PR #51, opened 2021-12-09
Add autogenerate_from_command field to Widget contribution

### LINKED ISSUE #47: migration support for napari_experimental_provide_function hooks
### Description

I discussed this in separate conversations w @tlambert03, @sofroniewn and in the plugin working group meeting.

One of the nice things about migrating these is that it makes it simple for people to integrate functions like these into napari without adding dependencies to their package (no npe2, magicgui, napari etc...just the manifest file)

The proposal is to migrate napari_experimental_provide_function hooks to be npe2 widget contributions. It sounds like something like this might work:

```yaml
name: myplugin
contributions:
  commands:
    - id: myplugin.func
      python_name: myplugin:my_typed_function
      title: Open dock widget for myplugin function
  widgets:
    - command: myplugin.func
      name: Functionality
      type: magicgui                #<-- new field 
```

Implementation should cover:
- [x] schema support
- [x] npe2 conversion tool support
- [ ] npe2 doc update
- [ ] npe2 cookiecutter update

tlambert03:
closes #47 

Note, I opted not to use `type: magicgui`, since I think it could be a little confusing.  Someone might use magicgui on their own, and think that we're asking them to tell us whether they used magicgui or Qt to generate their widget.  `autogenerate_from_command` is a bit verbose, but hard to confuse :)

side note: I had to modify a couple things about the plugin ContributionIndex, there was a bug where it was retaining entries across instances


nclack (approved):
fantastic. opened napari/napari#3789 for the display_name change.

sofroniewn:
> `autogenerate_from_command` is a bit verbose, but hard to confuse :)

that's fine, but i would have suggested `autogenerate` as being clear in context and a little more friendly, unless you think we'd ever have other `autogenerate_from_....` fields, but those are likely to be incompatible with a bool on `autogenerate_from_command`

PR #52, opened 2021-12-09
[WIP] Test conversion for all plugins

tlambert03:
This is an experiment to see if we can determine how successful the conversion tool will be across all known plugins.


npe2/manifest/commands.py
```diff
@@ -9,12 +9,13 @@
...
 
 _identifier = "([a-zA-Z_][a-zA-Z_0-9]+)"
+_identifier_plus_dash = "([a-zA-Z_][a-zA-Z_0-9-]+)"
 
 # how do we deal with keywords ?
 # do we try to validate ? Or do we just
 # assume users won't try to create a command named
 # `npe2_tester.False.if.for.in` ?
-_dotted_name = f"(({_identifier}\\.)*{_identifier})"
+_dotted_name = f"(({_identifier_plus_dash}\\.)*{_identifier_plus_dash})"
```
nclack, line 18: i did this in a conflicting way in #51.  the way you did it here looks better.

nclack (approved):
very nice

tlambert03:
> but I wish there a way to surface these checks separately from the required CI checks.

If you mean you'd like these not to run all the time on CI... absolutely.  I temporarily set the trigger to be "on push", but just made it to "workflow_dispatch" so it will only run on demand.

tlambert03:
need some stuff here, so merging.  This workflow isn't "active" yet.  just on manual trigger.

Issue #59, opened 2021-12-12
Issues with npe2 convert


jni:
I ran npe2 convert on affinder at https://github.com/jni/affinder/tree/1b35d64bb8eb57b842ba15162ccbd57048f04280

(Side note: I expected that I would run npe2 convert on my *code*, not on installed plugins. `npe2 convert .` gave me a meaningful error but I kinda would have wanted it to just look at a repo rather than an installed package. Maybe that's harder though.)

1. Running `npe2 convert affinder > napari.yaml` gives a warning, `Invalid schema 'napari:builtins.yaml'`, and no other feedback. It actually succeeded, so a success message should probably also be printed on stderr?
2. npe2 convert puts the napari.yaml in the current directory, but it actually needs to go in the package directory and get added to the manifest, right? That can probably be done automatically?

tlambert03:
> I kinda would have wanted it to just look at a repo rather than an installed package

Yeah, that's kind of the whole point we're moving to npe2 😂 ... because we _had_ to install/import plugins in npe1 to see what they offer. 

> npe2 convert puts the napari.yaml in the current directory

It puts it wherever you tell it to using the `-o` flag, or, here, using the `>`. 
Point taken though.  Currently, all npe2 convert does is generate the manifest, we still need to modify setup.cfg... and that's the point where it really matters where the file ends up.

PR #60, opened 2021-12-12
Make `npe2 convert` modify a repository

tlambert03:
This is to address https://github.com/napari/npe2/issues/59

currently `npe2 convert` expects an npe1 plugin name, and outputs a manifest to stdout, or to whatever name the user gives for the `--out` flag.

That still leaves the user with some work to do: putting the file in the top level module, and modifying entry points in `setup_cfg`.  this PR turns `convert` into an actual repo conversion.  To get the old behavior of just printing the manifest use `npe2 convert . --just-print`  (but note, it still takes a patch now, and not a package name).

This doesn't eliminate the need for the path to be an _installed_ package in the current repository.  So this will only work if the path being provided points to a package installed in the current environment.  It's hard to get around this, since npe1 _has_ to import in order todetect all the contributions.


nclack:
@tlambert03 What's the status here?  It looks like there's a related commit on main that should go here.

I really love this.  It's very convenient and seems to work pretty well on my system. 

some notes from running it:
 - it feels like it should be an opt-in/safe-mode before files are written. something like `git clean -n`
 - the old lookup-by-name option might be nice to have around?  I can't think of why though. so just mentioning.

tlambert03:
> @tlambert03 What's the status here? It looks like there's a related commit on main that should go here.

it's close 

> * it feels like it should be an opt-in/safe-mode before files are written. something like `git clean -n`

by default?  or only if you add the dry-run flag?  (my preference is to modify by default.  It's only adding a new file, and making a one-line modification to setup.cfg.  I can add something to the cli help string (i think the word "convert" does imply that we're modifying too)

> * the old lookup-by-name option might be nice to have around?  I can't think of why though. so just mentioning.

agreed, that's one thing I'm trying to bring back before merge

nclack:
> add the dry-run flag

this one. basically just print out the files that will get modified.

tlambert03:
ok, I think this is ready enough to merge.  the workflow still shows that a number of plugins fail, but i think it's mostly on the plugin side of things.  turned off the workflow for now

tlambert03:
@nclack  feel free to merge when you want.  current cli is 

```
npe2 convert .  # turn current directory  into new style plugin (it MUST also be installed in current env)
npe2 convert some-package  # print out manifest for some-package
npe2 convert -n .  # dry run for convert
```

nclack:
Ran into a problem.  When make build and install the bdist of the converted package and try to run napari I get this:

```
napari.manifest -> 'napari_animation:napari.yaml' could not be imported: Could not find file 'napari.yaml' in module 'napari_animation'
```

It looks like napari.yaml is not begin placed in the root module folder. It looks like setup.cfg is also missing 

```ini
[options.package_data]
napari_animation =
    napari.yaml
```

If I move the `napari.yaml` into `napari_animation` and add the package_data thing, then it works.

tlambert03:
Ah right! Good catch thanks, will fix

tlambert03:
fixed, will merge after tests pass

PR #61, opened 2021-12-13
Start to make command APIs clearer

tlambert03:
the parallel of hook_specifications.py for npe2.  Not sure yet were to surface/use these things, but we definitely need them somewhere.

note, this pr also makes npe2.types public

### DIFF
--- a/npe2/_types.py
+++ b/npe2/_types.py
@@ -54,11 +54,13 @@ ReaderGetter = Callable[[Union[str, List[str]]], Optional[ReaderFunction]]
 # SampleDataGenerator.command must point to a SampleDataCreator
 SampleDataCreator = Callable[..., List[LayerData]]
 
-# WriterContribution.command must point to a WriterFunction
-# Writers that take at most one layer must provide a SingleWriterFunction command.
-# Otherwise, they must provide a MultiWriterFunction.
-# where the number of layers they take is defined as
-# n = sum(ltc.max() for ltc in WriterContribution.layer_type_constraints())
+# WriterContribution.command must point to a WriterFunction.
+# Currently, two calling conventions are supported for writers: single-layer and
+# multi-layer writers. When at most one layer can be matched by a writer, it
+# must use the single-layer convention. Otherwise, the multi-layer convention
+# must be used. WriterFunctions are only called when the set of layers to be
+# written satisfies a set of layer-type constraints specified by
+# WriterContribution.layer_types.layer_type_constraints().
 SingleWriterFunction = Callable[[str, DataType, Metadata], List[str]]
 MultiWriterFunction = Callable[[str, List[FullLayerData]], List[str]]
 WriterFunction = Union[SingleWriterFunction, MultiWriterFunction]
--- a/npe2/_types.py
+++ b/npe2/_types.py
@@ -1,12 +1,21 @@
 from pathlib import Path
-from typing import TYPE_CHECKING, Callable, Dict, List, Sequence, Tuple, Union
+from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Sequence, Tuple, Union
 
 from typing_extensions import Literal, Protocol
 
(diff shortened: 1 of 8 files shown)


npe2/_types.py
```diff
@@ -31,9 +40,28 @@ def __array__(self) -> "np.ndarray":
...
+
+# SampleDataGenerator.command must point to a SampleDataCreator
+SampleDataCreator = Callable[..., List[LayerData]]
+
+# WriterContribution.command must point to a WriterFunction
+# Writers that take at most one layer must provide a SingleWriterFunction command.
+# Otherwise, they must provide a MultiWriterFunction.
+# where the number of layers they take is defined as
+# n = sum(ltc.max() for ltc in WriterContribution.layer_type_constraints())
+# @nclack, please check ... and let's see if we can make that clearer in the docstrings
```
tlambert03, line 62: @nclack 👀

nclack (approved):
Thanks for doing this! I tried with the writer comment...

npe2/_types.py
```diff
@@ -31,9 +40,28 @@ def __array__(self) -> "np.ndarray":
 DataType = Union[ArrayLike, Sequence[ArrayLike]]
 FullLayerData = Tuple[DataType, Metadata, LayerName]
 LayerData = Union[Tuple[DataType], Tuple[DataType, Metadata], FullLayerData]
```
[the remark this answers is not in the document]
tlambert03: sure, happy to do it.  I don't necessarily intend for this to be the document.  (wherever we put those efforts should be done in such a way that it can also go on the website, easily)

tlambert03:
> I tried with the writer comment...

didn't follow this.  did you type something somewhere?

npe2/_types.py
```diff
@@ -31,9 +40,28 @@ def __array__(self) -> "np.ndarray":
...
+ReaderGetter = Callable[[Union[str, List[str]]], Optional[ReaderFunction]]
+
+# SampleDataGenerator.command must point to a SampleDataCreator
+SampleDataCreator = Callable[..., List[LayerData]]
+
+# WriterContribution.command must point to a WriterFunction
+# Writers that take at most one layer must provide a SingleWriterFunction command.
+# Otherwise, they must provide a MultiWriterFunction.
+# where the number of layers they take is defined as
+# n = sum(ltc.max() for ltc in WriterContribution.layer_type_constraints())
```
nclack, line 61:
```suggestion
# WriterContribution.command must point to a WriterFunction.
# Currently, two calling conventions are supported for writers: single-layer and
# multi-layer writers. When at most one layer can be matched by a writer, it
# must use the single-layer convention. Otherwise, the multi-layer convention
# must be used. WriterFunctions are only called when the set of layers to be
# written satisfies a set of layer-type constraints specified by
# WriterContribution.layer_types.layer_type_constraints().
```

nclack: try 2

npe2/_types.py
nclack:
```suggestion
```

npe2/_types.py
tlambert03: thx, merged.  i think what I was really looking for though, is something more explicit than "When at most one layer can be matched by a writer, it must use the single-layer convention. Otherwise, the multi-layer convention must be used.".  This docstring _still_ requires me to go elsewhere to figure it out.  it references `WriterContribution.layer_types.layer_type_constraints`, which also doesn't have a docstring... so I have to dig deeper. 

`n = sum(ltc.max() for ltc in WriterContribution.layer_type_constraints())` was a bit better, since it tells me _exactly_ what it means to match more than one layer ... but there too, I have to go through `LayerTypeConstraint.max` and `LayerTypeConstraint.bounds` before I finally get to something that tells me what the _manifest_ needs to do to indicate these things.  Let's do this in a follow up though

PR #62, opened 2021-12-14
Remove semver dependency, vendor small portion

tlambert03:
mostly just need the regex.  but it's kind of nice to have a class with comparisons, so implemented that too


nclack (approved):
lgtm thanks!

PR #63, opened 2021-12-14
Autogenerate contribution docs from schema

tlambert03:
@nclack, here's rough example of how we could autogenerate nice looking markdown, while still interjecting our own prose, using jinja.  run `pip install Jinja2` then `python docs/render.py` and look in `docs/_build`

examples:

Screen Shot 2021-12-14 at 5 03 44 PM

Screen Shot 2021-12-14 at 5 04 08 PM

we can move the prose you added in the docs PR to the class docstrings and field descriptions to fill in the details.
Then, in napari, to generate the docs, it could clone this repo and run `python docs/render.py` or whatever


nclack:
thanks!  I was imagining fiddling with sphinx but this is much better.  It'll be a bit before I look into it ;)

tlambert03:
ok @nclack ... I think this is ready for a closer look.  This is more or less how I'd like the contributions page to be rendered, and it's pulling everything from the docstrings and fields themselves.  (The examples are pulled from the manifest at docs/example_manifest)

I temporarily [added the rendered result here](https://github.com/napari/npe2/blob/9cd6b1add05d2c3737bc8d16d55f79b0c21a2d58/docs/_build/contributions.md) so you can see how it ends up.

docs/example_manfest.yaml
```diff
@@ -0,0 +1,57 @@
+name: my-plugin
```
nclack, line 1: thinking about renaming this to example-plugin

docs/render.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def main():
+    env = Environment(loader=PackageLoader("docs"), autoescape=select_autoescape())
+    _BUILD.mkdir(exist_ok=True, parents=True)
+    suffix = ".md"
+    for t in TEMPLATES.glob("*.jinja"):
+        template = env.get_template(t.name)
+        context = {"contributions": _get_contributions()}
+        (_BUILD / f"{t.stem}{suffix}").write_text(template.render(context))
```
nclack, line 98: Two small things here:
- I'd like to rename the templates to something like `contributions.md.jinja`. This gets rid of the need for `suffix` here.
- I always appreciate seeing the modified paths get printed out.

```suggestion
def _log(path: Path) -> Path:
    print(path)
    return path


def main():
    env = Environment(loader=PackageLoader("docs"), autoescape=select_autoescape())
    _BUILD.mkdir(exist_ok=True, parents=True)
    for t in TEMPLATES.glob("*.jinja"):
        template = env.get_template(t.name)
        context = {"contributions": _get_contributions()}
        _log(_BUILD / f"{t.stem}").write_text(template.render(context))
```

nclack (commented):
I'm still looking at this but some initial thoughts:

If the templates get named something like `contributions.md.jinja` then I get good syntax highlighting in vscode via the Better Jinja extension. It _also_ means that we don't need to append the suffix in render which I think makes the code a little more flexible.

I really love the example manifest. I thought "example plugin" might be better than "my plugin".

tlambert03:
> If the templates get named something like `contributions.md.jinja` then I get good syntax highlighting in vscode via the Better Jinja extension.

lovely!  that's much "better" than the extension I'd been using: https://marketplace.visualstudio.com/items?itemName=wholroyd.jinja :)

docs/example_manfest.yaml
tlambert03: renamed

PR #67, opened 2021-12-17
convert adds napari.yml to package_data

nclack:
`npe2 convert` wasn't adding the manifest to `options.package_data`.  

This PR aims to do that.

I tested it by doing

```
npe2 convert .
python setup.py bdist_wheel | grep yaml # can see the file in the output
pip install dist/<name of whl>
napari -vv
```

And checking the plugin is there/works.


nclack:
nevermind!

PR #68, opened 2021-12-17
Replace entry_point with activate/deactive function

tlambert03:
as it sounds, it top level manifest

### DIFF
--- a/npe2/manifest/commands.py
+++ b/npe2/manifest/commands.py
@@ -1,22 +1,15 @@
-import re
 from textwrap import dedent
 from typing import TYPE_CHECKING, Any, Optional
 
 from pydantic import BaseModel, Extra, Field, validator
+
+from . import _validators
 
 if TYPE_CHECKING:
     from .._command_registry import CommandRegistry
 
 _distname = "([a-zA-Z_][a-zA-Z0-9_-]+)"
 _identifier = "([a-zA-Z_][a-zA-Z_0-9]+)"
-_identifier_plus_dash = "([a-zA-Z_][a-zA-Z_0-9-]+)"
-
-# how do we deal with keywords ?
-# do we try to validate ? Or do we just
-# assume users won't try to create a command named
-# `npe2_tester.False.if.for.in` ?
-_dotted_name = f"(({_identifier_plus_dash}\\.)*{_identifier_plus_dash})"
-_python_name_pattern = re.compile(f"^{_dotted_name}:{_dotted_name}$")
 
 
 class CommandContribution(BaseModel):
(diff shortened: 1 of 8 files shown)


npe2/manifest/schema.py
```diff
@@ -101,16 +102,25 @@ class PluginManifest(BaseModel):
     # the actual mechanism/consumption of plugin information) independently
     # of napari itself
 
-    # TODO: refactor entry_point to binding points for activate,deactivate
-    # TODO: Point to activate function
-    # TODO: Point to deactivate function
-
-    # The module that has the activate() function
-    entry_point: Optional[str] = Field(
+    activate_function: Optional[str] = Field(
```
nclack, line 105: I'm not a fan of the "function" in the name.  On suggestion might be:

```suggestion
    on_activate: Optional[str] = Field(
```

npe2/manifest/schema.py
```diff
@@ -101,16 +102,25 @@ class PluginManifest(BaseModel):
...
+        _validators.python_name
+    )
+    deactivate_function: Optional[str] = Field(
         default=None,
-        description="The extension entry point. This should be a fully "
-        "qualified module string. e.g. `foo.bar.baz` for a module containing "
-        "the plugin's activate() function.",
+        description="Fully qualified python path to a function that will be called "
+        "when a user deactivates a plugin (e.g. my_plugin._some_module:deactivate). "
+        "This is optional, but may be used to perform any plugin cleanup.",
```
nclack, line 120:
```suggestion
        "This is optional, and may be used to perform any plugin cleanup.",
```

npe2/manifest/commands.py
```diff
@@ -1,22 +1,15 @@
...
 
 from pydantic import BaseModel, Extra, Field, validator
 
+from . import _validators
+
 if TYPE_CHECKING:
     from .._command_registry import CommandRegistry
 
 _distname = "([a-zA-Z_][a-zA-Z0-9_-]+)"
 _identifier = "([a-zA-Z_][a-zA-Z_0-9]+)"
```
nclack, line 12: This is more of an observation than anything, but I really like the cleanup you did w _validators but noticed these didn't move in.

nclack (commented):
I left a few comments, but didn't go over it in detail yet.

npe2/manifest/schema.py
tlambert03: thanks, I was also so-so on that.  I considered `on_activate` in my head too, so that's an easy one

npe2/manifest/commands.py
tlambert03: can do

npe2/manifest/schema.py
tlambert03: renamed

nclack (approved):
looks good! I made pr's to napari and napari-svg to remove the entry_point field so that they'll be compatible with this change.

PR #69, opened 2021-12-17
change engine to schema_version

### LINKED ISSUE #64: reconsidering `engine` name
hey @nclack ... I was looking at the vscode docs, and realized that they use "engine" to mean the version of _vscode_ that the package is targeting (not the version of the manifest).  On its own, I don't think "engine" is immediately obvious (though, perhaps someone could figure it out because it's called np**e**).  But, what would you think of using `schema_version` instead?  Hard to get that wrong

tlambert03:
closes #64 

renames `manifest.engine` to `manifest.schema_version`


nclack (approved):
thanks! looks good.

PR #70, opened 2021-12-17
Bring test coverage to 100%, reorganize tests

tlambert03:
This brings complete test coverage to npe2, (and requires 100% for future PRs).

I tried to touch as little of the actual code as possible, but inevitably needed to shuffle a few things around (like validators, etc...)  hopefully it's not too hard to review.  But, it's all tested! 😂 

you'll see a lot of `# pragma: no cover` ... mostly they are on exceptions and loop continuations, those are fine with me (mocking test cases that specifically fail those conditions just to see the exception get raised is not terribly useful, and each one indicates that a human has considered that line).  I'm fine with using `# pragma: no cover`  going forward when it feels like a "trivial" thing that


tlambert03:
@nclack, let me know if you'd like time to review this PR.  I know we're in a holiday stretch here, so I expect you won't be as active as usual.  This PR has the potential to conflict with anything else I work on though, so I'd prefer to get it in sooner than later. (as mentioned, it really doesn't touch much of the logic in the code, just completes coverage)

npe2/_from_npe1.py
```diff
@@ -481,7 +485,7 @@ def _write_new_setup_cfg_ep(info: PluginPackage, mf_name: str):
     mf_path = f"{info.top_module}:{mf_name}"
     new_ep = f"\n{info.package_name} = {mf_path}"
     if "options.entry_points" not in p.sections():
-        p.add_section("options.entry_points")
+        p.add_section("options.entry_points")  # pragma: no cover
     p.set("options.entry_points", NPE2_EP, new_ep)
     if "options.package_data" not in p.sections():
         p.add_section("options.package_data")
```
nclack, line 491:
```suggestion
    def _mksection(p,name:str):
        if name not in p.sections():
            p.add_section(name)
    _mksection(p,"options.entry_points")
    p.set("options.entry_points", NPE2_EP, new_ep)
    _mksection(p,"opstions.package_data")
```

npe2/_from_npe1.py
```diff
@@ -481,7 +485,7 @@ def _write_new_setup_cfg_ep(info: PluginPackage, mf_name: str):
     mf_path = f"{info.top_module}:{mf_name}"
     new_ep = f"\n{info.package_name} = {mf_path}"
     if "options.entry_points" not in p.sections():
-        p.add_section("options.entry_points")
+        p.add_section("options.entry_points")  # pragma: no cover
```
nclack, line 488: This is logic that will get in the the "real world". So flagging this as a maybe-don't-pragma-this-one.

It doesn't really need coverage. The use of `add_section` for `package_data` is hit below.

There's sort of a goofy way to cover both by factoring a method gets reused for lines on 487 and 491.

nclack: added a suggestion showing what I mean. I'm ok with the pragma. just pointing this out.

tests/sample/my_plugin-1.2.3.dist-info/METADATA
```diff
@@ -1,3 +1,6 @@
 Metadata-Version: 2.1
 Name: my_plugin
 Version: 1.2.3
+Summary: My napari plugin
+License: BSD-3
+Author: The Black Knight
```
nclack, line 6: spooky

npe2/_plugin_manager.py
```diff
@@ -313,7 +309,7 @@ def get_writer(
 
         for writer in self.iter_compatible_writers(layer_types):
             if plugin_name and not writer.command.startswith(plugin_name):
-                continue
+                continue  # pragma: no cover
```
nclack, line 312: This will get hit when a user has multiple single-image-layer writers installed (from different plugins) and is querying by plugin_name.

Right now I believe the testing is focused on environments with just one test plugin installed. Should we add another to cover this?

nclack (approved):
amazing.  I flagged a couple pragmas, but I don't feel strongly. It's so nice to have this.

npe2/_from_npe1.py
nclack: I left another review comment referencing this suggestion. I don't really think it's worth adding the function to get rid of the pragma...

npe2/_from_npe1.py
tlambert03: > This is logic that will get in the the "real world". 

actually, this conditional will probably _never_ hit.  The reason is that `_write_new_setup_cfg_ep` only gets called if the plugin is using `setup.cfg` (instead of `setup.py`), and if they don't already have an `entry_points` section in the `setup.cfg`, then they're effectively not a plugin anyway, and it wouldn't have made it to this point.  We could even just delete this conditional, but it seems safer to include, but not worth testing

npe2/_plugin_manager.py
tlambert03: the logic of simply continuing is simple enough.  But I refactored out the `continue` and the pragma with `if not plugin_name or writer.command.startswith(plugin_name):`

PR #73, opened 2021-12-20
split contributions/io into readers and writers

tlambert03:
`contributions/io.py`, unlike all other contributions, has two classes (readers and writers).  I frequently get tripped up by that when I'm searching files, and would like to standardize it to be like the rest


nclack (approved):
lgtm

PR #74, opened 2021-12-21
change sample plugin name from `my_plugin` to `my-plugin`

tlambert03:
To avoid confusion between packages and modules, this changes the name of the sample plugin that we test to `my-plugin` from `my_plugin` (which is the top module)


nclack (approved):
thanks!

PR #75, opened 2021-12-21
add `@npe.implements` decorators, for opt-in manifest validation and/or AST-based-generation

### LINKED ISSUE #66: Task: add decorator to aid in manifest construction/validation
It might be nice to add a couple decorators like this:

```py
import npe2

@npe2.widget(display_name="My Widget")
class MyWidget(QWidget):
    ...

@npe2.reader(filename_patterns=["*.tif"])
def read_tiff(path): ...
```

these would be no-op at runtime, but could allow this:

```
npe2 compile-manifest .
```

would would build a manifest for you out of the contributions... 

this could also be used in a pre-commit hook, to make sure that the manifest agrees (for instance, if you rename or move a function).  basically just an opt-in feature to make it easier to build a manifest and/or keep it in sync with your code

tlambert03:
This implements an `@npe.implements...` decorator that allows someone to tag a function as implementing a specific contribution.  

```py
# my_plugin/__init__.py
from npe2 import implements

@implements.reader(
    id="some_reader",
    title="Some Reader",
    filename_patterns=["*.fzy", "*.fzzy"],
    accepts_directories=True,
)
def get_reader(path: str):
    ...
```

creates:
```yaml
contributions:
  commands:
    - id: my-plugin.some_reader
      title: Some Reader
      python_name: my_plugin:get_reader
  readers:
    - command: my-plugin.some_reader
      filename_patterns: ["*.fzy", "*.fzzy"]
      accepts_directories: true
```

The benefits are:

1. it could be used to autogenerate a manifest (there is a test included that confirms the decorated module yields the same manifest contributions as the manually created one)
2. it could be used to ensure that the manifest and the code never go out of sync: the decorators live with the python and  have negligible runtime-cost
3. The `compile` command is entirely static: it's purely based on the AST.  so it's very fast and could be used for pre-commit hooks
4. Conceptually (though I'm not proposing or suggesting this), this could eliminate the need for user-generated manifests entirely for simple plugins.  The user could either add an `npe2 compile` step to their package build process, or somehow indicate to napari the files that it should _(statically)_ analyze once after install.
  
Closes #66


tlambert03:
side note: I started on this because I wanted a way to "mark" a given function as implementing a given contribution when autogenerating documentation.  So this would also be useful for that

tlambert03:
cc @ctrueden ... similar to what we had discussed.  curious to get your thoughts if you're so inclined

ctrueden:
@tlambert03 Nice! Seems like a great step forward to keeping plugin metadata in the same place as the plugin itself, to avoid action-at-a-distance type errors where the developer forgets to update the YAML metadata after changing plugin code.

About manifest autogeneration: were you thinking to support partial manual curation of YAML files, with the contributions section unioned with the compiled block? Or only all or nothing? I'm not familiar enough with the `napari.yaml` specification to be sure, but I assume there are other useful things that can be filled in that aren't amenable to autogeneration? So it would be important to synthesize the final manifest from the two different sources in that case, no?

tlambert03:
> About manifest autogeneration: were you thinking to support partial manual curation of YAML files, with the contributions section unioned with the compiled block? Or only all or nothing? I'm not familiar enough with the napari.yaml specification to be sure, but I assume there are other useful things that can be filled in that aren't amenable to autogeneration? So it would be important to synthesize the final manifest from the two different sources in that case, no?

yes, there are few things that are purely static at the moment, like theme contributions, and probably menus/submenus contributions.  So this would most likely be partial/unioned with the manually edited stuff.

That said, for a simple widget or reader plugin (which probably constitutes 80-90% of the plugins), it probably _would_ be sufficient just to have the decorators (since we already glean a lot of the additional needed info from package metadata in `setup.cfg`)

nclack (commented):
This is very cool.  I really like the direction of making it easier to maintain code wrt the manifest.  If I were using, this would be my prefered way of setting up a plugin. I don't want to eliminate the manifest but reducing the amount of stuff in there and keeping the annotations close to the code is great.

I want to look at this more in detail but some initial thoughts/observations:

Would it be possible to type check the decorated calls?

I'm trying to think through what happens as the spec evolves - if we deprecate a contribution.  `compile()` could be useful for forward migations. Might want to namespace the decorators somehow.

npe2/implements.py
```diff
@@ -0,0 +1,196 @@
...
+# @overload
+# def _marker(__obj: Literal[None]=None, **kwargs) -> Callable[[T], T]: ...
+# @overload
+# def _marker(__obj: T, **kwargs) -> T: ...
+# # fmt: on
+# def _marker(__obj: Optional[T] = None, **kwargs) -> Union[T, Callable[[T], T]]:
+#     def _inner(obj: T) -> T:
+#         return obj
+
+#     return _inner if __obj is None else __obj
```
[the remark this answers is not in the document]
tlambert03: wip: was working on typing for the decorated functions... still needs to be done.  (typing decorators is a bit tricky).  I'll probably make a pyi file for this particular module

tlambert03:
> Would it be possible to type check the decorated calls?

do you mean can users perform type checking on their own functions if they use the decorator?  or can we type check something about the decorator?  (both yes, for sure ... and I'll do more before this goes in but maybe clarify which one you were thinking about)

> I'm trying to think through what happens as the spec evolves - if we deprecate a contribution. compile() could be useful for forward migations.

indeed!

> Might want to namespace the decorators somehow.

elaborate?  I was thinking that `implements` _is_ the namespace... but I suspect i'm missing the point

jni:
Eh. tbh I don't actually love this. I think for packages that are *only* napari plugins this makes some sense (though it duplicates some information, which I also don't love). But one design principle that's always guided me in thinking about napari plugins is that whole "napari headless is just Python", and a corollary, "napari plugins without napari should just be Python modules". As a thought experiment, would I decorate all scikit-image functions with something like this? Heck no. But would I add a napari.yaml to `skimage/`? Quite possibly!

So, I'm totally fine with it as **opt-in**, but I just want to guard against this becoming the default or even preferred way of implementing plugins.

Another thought: if we are doing `npe2 compile` to make sure that the napari.yaml is up to date, isn't this equivalent to having `npe2 validate` check the yaml against the Python module?

tlambert03:
> I think for packages that are only napari plugins this makes some sense

Yep, which, let's be honest, is probably over 95% of the plugins.  Even with our attempts to make it work with any package, we've seen multiple packages _extract_ their plugin into a new package (ome-zarr, stardist, etc)...

But yes, I definitely am not proposing we not support manual manifests.  So the scikit-image case will certainly be kept in mind.  but if we're being honest/observant, I think we need to admit that that's very much an edge case, and for the 1 widget, or 1 reader, plugins: this obviates the need for a two file solution. 

Yes, it's also possible to reverse check: going from manifest to python name.

jni:
> for the 1 widget, or 1 reader, plugins: this obviates the need for a two file solution.

wait, it doesn't, does it? napari.yaml still needs to exist, it'll just be autogenerated?

> we've seen multiple packages extract their plugin into a new package (ome-zarr, stardist, etc)...

Yes, I consider this a *symptom* that our plugin system was too heavyweight (requiring imports and runtime dependencies), not a reason to give up on the ideal. ie my reading of the situation is that in the napari.yaml world this might not have happened. In the npe.implements world, we are back in the previous situation that developers have tried to avoid.

tlambert03:
> wait, it doesn't, does it? napari.yaml still needs to exist, it'll just be autogenerated?

yes, but it could essentially be entirely invisible to the plugin developer in some cases.  As in, in certain cases, they would never look at it or touch it.  It would be an implementation detail of napari discovery.

> I consider this a symptom that our plugin system was too heavyweight

I do agree with this!  and i definitely want to re-emphasize my point in the first post that I'm not suggesting this be the mandatory design.

> In the npe.implements world, we are back in the previous situation that developers have tried to avoid.

except that these decorators are purely statically discovered.  You point to a file path, and _without import_, these are found (that really was the primary "niceness" of this PR)

--- 

I'm currently working on the opposite validation direction as discussed above (given manifest, import code and check python names), but can you let me know if you're feeling like "please don't merge this", or if you're feeling like "i probably won't ever use that, but go ahead and experiment"

jni:
> yes, but it could essentially be entirely invisible to the plugin developer in some cases. As in, in certain cases, they would never look at it or touch it. It would be an implementation detail of napari discovery.

Will the napari.yaml be generated at wheel/sdist generation time then? If so I must admit that's cool. ;)

> I'm not suggesting this be the mandatory design.

See this phrasing worries me. :joy: As I mentioned, I'd like this to be *the alternative*, rather than the *primary* pattern. "Here's how you make a plugin (napari.yaml); btw, hey, are you making a quick and dirty plugin based on a few napari-specific functions? Here's a shortcut!"

> except that these decorators are purely statically discovered. You point to a file path, and without import, these are found (that really was the primary "niceness" of this PR, with this visitor)

This is all very cool, but it's an implementation detail of npe2, right? That is, from a developer perspective, they have an actual import in their code and this import means that they need to depend on npe2?

> I'm currently working on the opposite validation direction as discussed above

:pray: 

> can you let me know if you're feeling like "please don't merge this", or if you're feeling like "i probably won't ever use that, but go ahead and experiment"

The latter, :blush: with the caveat that I mentioned above: I wouldn't want this to be The Blessed Way.

tlambert03:
> Will the napari.yaml be generated at wheel/sdist generation time then? If so I must admit that's cool. ;)

yep, that's the idea.  Or, even possibly: _after_ python package install, at first runtime (with a cached result that presumably never changes after install)

> See this phrasing worries me. 😂 

lol... yes yes, we're on the same page :joy:

> That is, from a developer perspective, they have an actual import in their code and this import means that they need to depend on npe2?

indeed! 

> I wouldn't want this to be The Blessed Way.

agreed!  I think it merits some further investigation (as does running automated manifest-to-python checks)

npe2/implements.py
```diff
@@ -0,0 +1,196 @@
...
+        kwargs["command"] = cmd.id
+        return kwargs
+
+    def _qualified_pyname(self, obj_name: str) -> str:
+        return f"{self.module_name}:{obj_name}"
+
+
+def compile(
+    path: Union[ModuleType, PathLike], plugin_name: str, module_name: str = ""
+) -> dict:
```
jni, line 183: I'd love to see docstrings for all these, even the private ones. As I'm super unfamiliar with all the patterns being used here, the code is very hard to parse. Docstrings could make it easier for newcomers to follow/fix/improve.

jni:
Good stuff. 😊 And I agree — after some experimentation you might turn me around on which way should be the most common. 😊

npe2/implements.py
tlambert03: will do!  this was "done enough" for a general concept review, but i'll fill out the details (type stubs and docs) before anything merges

tlambert03:
Marking this old PR ready for review.  After chatting with @DragaDoncila, I like the idea even more.  It isn't really exposed in any way yet, but would be nice to have it available to play with

DragaDoncila (approved):
I'm approving this purely based on tests passing and because we're not using it anywhere so it gives us the opportunity to play with it.

I have to confess I don't really understand the code. I think in the interest of maintenance and knowledge sharing we need a document that explains this approach. 

I know much of it requires understanding of AST parsing, which is outside the scope of this project, but some things would be nice to have explanations for e.g. how this is even working when the text "napari_hook_implementation" doesn't even appear in the entire `PluginModuleVisitor` class? Additionally if you know of a good intro tutorial/documentation for the AST, a link would be much appreciated.

Usage documentation would also be nice. None of this should necessarily hold up the merge though, as we are not exposing this anywhere just yet.

tlambert03:
> how this is even working when the text "napari_hook_implementation" doesn't even appear in the entire `PluginModuleVisitor` class? 

Thank you @DragaDoncila for punishing yourself through this :)  I apologize I didn't document it better, definitely necessary for this one.  Should have done before you reviewed, and will do now,.

To answer this question though, note that is _not_ looking for npe1 style decorators, instead it's looking for things decorated with stuff from `npe2.implements` as shown in the original post above.  And the way _that_ works all comes down to [these lines here](https://github.com/napari/npe2/pull/75/files#diff-4e002e69302a85e4d391bd4616b9f2e16a5a11d1ccae50c7a09e994014c29cebR114-R115):

```python
    def _find_decorators():
        ...
                    if self._names.get(call_name) == __name__:
                         self._store_contrib(call.func.attr, call.keywords, node.name)
```

What's happening there is that `self._names` are populated by [all the imports in the module](https://github.com/napari/npe2/pull/75/files#diff-4e002e69302a85e4d391bd4616b9f2e16a5a11d1ccae50c7a09e994014c29cebR87-R95).  So what ` if self._names.get(call_name) == __name__:` essentially means is "if a function is being decorated with something from `npe2.implements` (because `__name__` in that module is going to be `npe2.implements`)

> Additionally if you know of a good intro tutorial/documentation for the AST, a link would be much appreciated.

I like https://greentreesnakes.readthedocs.io/ !
The AST is definitely intimidating, but also kinda awesome-sauce and empowering.  highly recommend that page.


In any case, i fully agree, this module needs more inline comments and general docs.  Will do!

tlambert03:
ok, here's the class docstring I added to the ModuleVisitor (pasted here for convenience).  Sorry again for not adding this before you looked!

-----
This visitor will find all the contributions in a module and store them in
`contribution_points`.  It works as follows:

1. Visit all `Import` and `ImportFrom` nodes in the module, storing their import
   names in `_names` so we can look them up later.  For example, if the module
   had the line `from npe2 import implements as impls` at the top, then the
   `visit_ImportFrom` method would add the entry:
   `self._names['impls'] = 'npe2.implements'`
   This way, we know that an `@impls` found later in the module is referring to
   `npe2.implements`.
2. Visit all `FunctionDef` and `ClassDef` nodes in the module and check their
   decorators with `_find_decorators`
3. In `_find_decorators` we check to see if the name of any of the decorators
   resolves to the something from *this* module (i.e., if it's being decorated with
   something from `npe2.implements`).  If it is, then we call `_store_contrib`...
4. `_store_contrib` first calls `_store_command` which does the job of storing the
   fully-resolved `python_name` for the function being decorated, and creates a
   CommandContribution. `_store_contrib` will then create the appropriate
   contribution type (e.g. `npe2.implements.reader` will instantiate a
   `ReaderContribution`), set its `command` entry to the `id` of the just-created
   `CommandContribution`, then store it in `contribution_points`.
5. When the visitor is finished, we can create an instance of `ContributionPoints`
   using `ContributionPoints(**visitor.contribution_points)`, then add it to a
   manifest.

tlambert03:
@nclack ... I know you're very busy and this probably feels daunting.  I added a bunch of docs and walked through it with @draga.  this has sat for a while now so I'm inclined to merge (probably tomorrow), but let me know if you want a crack at it

nclack (approved):
looks great. I have some questions, but I think they can get answered in #186 etc

It looks like the workflow will be:
1. plugin developer adds the decorators
2. they use `compile` (from #186) to generate the manifest etc to make their plugin discoverable.

It looks like they'd need to add `npe2` to their plugin dependencies. Should `implements` be a separate pypi package?

Maybe related, it would be nice for folks to declare the schema version they're targeting with the decorators. The `compile` step should be able to handle migrations, but it will need some idea of the source schema version.

tlambert03:
> Maybe related, it would be nice for folks to declare the schema version they're targeting with the decorators. The compile step should be able to handle migrations, but it will need some idea of the source schema version.

they can still have a schema for things that aren’t command contributions (like display name, themes, etc…) so they would just have a regular manifest with far fewer contributions, and it would merge at compile time

> Should implements be a separate pypi package?

yeah it’s a good question.  But one I’m gonna punt on for now :). It’s not a terribly big dependency, and whatever does the compiling would need to depend on npe2 (but yes, i can imagine a package that just has the deocsrators, and npe2 would go in build_requires)

PR #76, opened 2021-12-21
Add option to validate python_name imports

tlambert03:
as mentioned/requested by @jni in the conversation in #75 

This adds a method `PluginManifest.validate_imports` and a new flag to the cli `npe2 validate --imports` that checks to see that all of the `python_names` used in the manifest are importable

this also has the very nice side effect of making `npe2.manifest.utils.import_python_name` the _only_ place in the code where we import things from packages

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,3 +104,4 @@ ENV/
 # IDE settings
 .vscode/
 npe2/_version.py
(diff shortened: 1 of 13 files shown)


nclack (approved):
This looks great!

PR #78, opened 2022-01-07
Move some _docs files

tlambert03:
This is in preparation of a PR to napari.   The previous structure made it a bit hard to gitignore the certain things over in napari.  This moves all of the static content (That doesn't need jinja) over to napari, where most people will probably look to edit it.

will follow up with a PR there afterwards


tlambert03:
@nclack, lemme know if you want time with this one.  It mostly just removes some files that are going to be added on the napari side.  but also adds a `make docs` command and some little conveniences

nclack (approved):
sorry slow. Approving with two concerns:

* I regret that this makes it so npe2 doesn't build its own standalone set of docs. 
* I feel like some of the moved files really should live in npe2 because they describes concepts specific to npe2. 

That said I don't think these should hold up the PR. Just mentioning as input.

tlambert03:
> I regret that this makes it so npe2 doesn't build its own standalone set of docs.

did you try running `make docs` here?  it _can_ build a standalone subset of the docs, but it clones a subset of the napari docs when doing so.  

Basically, we have to choose what goes where, and we can't duplicate stuff, since that would be silly.  We either put as much as possible over in napari, or as much as possible here.  With all the recent discussion about moving the tutorials and everything possible over to napari.org, I figured telling people that they have to go to `npe2` just to edit a static part of the plugin docs would go against that motivation.  So, this makes `npe2` only aware of the schema specific stuff that it has to test to stay in sync.  But it provides a convenience to test all this here running `make docs`

> I feel like some of the moved files really should live in npe2 because they describes concepts specific to npe2.

See above.  I'm not totally opposed to it, but I think that the community might be given all the recent discussion.  If we want to do that, I'd say that it should either _all_ live here (including the old section about npe1) or it should _all_ live there.  From an end-user's perspective, it feels more like "napari" docs, and they would probably go to the napari repo to ask for or suggest a change, which is why that is also a valid candidate (even if it's describing things that are _implemented_ by npe2)

Makefile
```diff
@@ -0,0 +1,11 @@
+.PHONY: docs
+
+docs:
+	rm -rf docs/plugins
+	python _docs/render.py
+	rm -rf ./napari
+	git clone --depth 1 --filter=blob:none --sparse -b update-plugin-docs https://github.com/tlambert03/napari.git
```
nclack, line 7:
```suggestion
        git clone --depth 1 --filter=blob:none --sparse -b update-plugin-docs https://github.com/napari/napari.git
```

nclack:
> did you try running make docs here?

I'm a bonehead. I was building manually. nvm!

Makefile
tlambert03: that won't work because napari doesn't have the update-plugin-docs branch.  this needs to be changed after the update-plugin-docs PR gets merged.

tlambert03:
gonna merge 

> (added a suggestion to change the repo in the makefile)

that can't be done because the branch being checked out isn't in the napari repo.  will have to come back here and change after the napari PR merge

Issue #50, opened 2021-12-09
complete IDE schema autocompletion tools


tlambert03:
Another good thing to do before release is to get our schema (once we finish cleanup and declare an initial version) onto https://www.schemastore.org/json/, where various vscode (e.g.) plugins can find it to provide autocompletion and hinting and stuff if someone is editing `napari.yaml`

nclack:
Schema is on schemastore so closing.

PR #83, opened 2022-01-12
add schema.json to repo root

tlambert03:
This adds a schema.json file to the root of the repo.  This will make it easier to accomplish #50, and actually should make it possible to put the entire jinja docs templating over in the napari repo (since we'd no longer need to import npe2 to get the schema)

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -1,0 +1,17 @@
+from pathlib import Path
+
+from npe2 import PluginManifest
+
+ROOT = Path(__file__).parent.parent
+
+
+def test_schema_current():
+    schema_string = PluginManifest.schema_json(indent=2) + "\n"
+    file = ROOT / "schema.json"
+    mismatch = bool(file.exists() and file.read_text() != schema_string)
+    file.write_text(schema_string)
+    if mismatch:
+        raise AssertionError(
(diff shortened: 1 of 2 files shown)


tlambert03:
@nclack, any opposition to this?  or thoughts on where to put it what to name it? 

could just put `schema.json` in the root dir, as done here.  Or could make a `schema` directory and put versioned copies in there.  If we do the latter, we should probably also put a "latest.json" in there so that the schemastore link can be permanent

nclack:
We [discussed ](https://hackmd.io/5EFaLRNlTxuF6tnYT1EQMw) in the plugin working group.  

Just noting the possible approach we identified for linking from schemastore: create a github release on deploy and link to the latest schema via the release link.

tests/test_schema.py
```diff
@@ -0,0 +1,17 @@
+from pathlib import Path
+
+from npe2 import PluginManifest
+
+ROOT = Path(__file__).parent.parent
+
+
+def test_schema_current():
```
nclack, line 8: could you add a comment describing the test's intent

nclack (approved):
Super cool to play around with. 

It might be worth documenting how people can take advantage of this...maybe a note in the readme or manifest guide.

PR #86, opened 2022-01-16
Add NPE1Shim, to convert npe1 plugins to the npe2 interface

tlambert03:
Still experimenting, but this PR would 
a) discover npe1 plugins along with npe2 during the discovery process
b) have some sort of way to generate a manifest from the entrypoints (using same logic as `npe2 convert`)
c) cache that manifest for use on later launches

this could be used to help us transition to npe2 being the only plugin manager


nclack:
There was some good discussion in the plugin working group meeting. See [notes](https://hackmd.io/XyX9yrQSRlWV1c-QP4eX3A).

Most of the discussion was around the caching behavior.

We discussed disable/uninstall support being tied to this which I think is great but probably a follow up pr?

tlambert03:
OK @nclack, @Carreau, this is now ready to go.  Due to the many moving parts, this has unfortunately ballooned into a massive PR.  So  I broke it up into three PRs.  

1. #124 - a more realistic test npe1 plugin, and logic to deal with locally defined objects in hookimpl scopes
2. #125 - the shim object itself, discovery logic, and plugin_manager awareness
3. #126 - caching logic and utilities

each PR contains the previous one... and #126 should actually be identical to this one.  If you prefer, you can also simply review this one (and I would just close those PRs).

The whole thing can be tested in concert by checking out this branch (or the branch from #126) along with https://github.com/napari/napari/pull/4015

PR #87, opened 2022-01-23
Handle list of paths in iter_compatible_reader

### LINKED ISSUE #81: bug: npe2 readers not prepared for lists of paths
whether it should get handled on the napari side (before passing paths to `npe2`) or on the npe2 side, the npe2 reader in io_utils is not currently handling a list of strings properly.

with a contribution of:
```yaml
readers:
    ...
    filename_patterns: ['*.ext']
```

`iter_compatible_readers(['f1.ext', f2.ext'])` will  return `None`

a list of strings is what will be passed when "open as stack" is true.

ppwadhwa:
I made some changes to `iter_compatible_readers` in order to accept a list of paths.  This will now open stacks.  <s>Also,  npe2 was not being used for folders, and I made a change to handle those as well.</s>. I see that it did handle folders, I realized it was not using the npe2 reader I was testing with to read.  

Fixes #81 .

### DIFF
--- a/npe2/_plugin_manager.py
+++ b/npe2/_plugin_manager.py
@@ -206,7 +206,7 @@ class PluginManager:
 
         is_directory = False
         if not isinstance(paths, list):
-            path = str(paths)
+            path = str(paths)  # needed in order to pass pre-commit hook
             if os.path.isdir(path):
                 # This is a directory.  Get a list of all files in the directory.
                 is_directory = True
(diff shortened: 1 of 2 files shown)


ppwadhwa:
Well---I guess I need to learn more about the directory reader.  Will fix this...

ppwadhwa:
OK.  So, I took out the stuff I had on folders.  It is not passing tests and I am not sure if I understand what needs to be done here anyway.  What I did notice, though, was that if my plugin reader can handle folders, it isn't getting there.  Its returning the images in a folder from the builtins reader.  

I am leaving that for another PR if we want to change that.  For now, this will handle lists of strings.

tlambert03 (approved):
thanks @ppwadhwa!

PR #92, opened 2022-02-02
set display_name to plugin name when empty

nclack:
By default `display_name` is the empty string. Client code has to coerce that to something reasonable (See napari/napari#4039).

It's a lot cleaner if we just do it here. This PR adds a rule to the `PluginManifest` root validator.


tlambert03 (commented):
thanks!

PR #97, opened 2022-02-09
make package meta hashable

tlambert03:


npe2/manifest/package_metadata.py
```diff
@@ -205,6 +205,9 @@ def from_dist_metadata(cls, meta: "email.message.Message") -> "PackageMetadata":
                 d[key] = value
         return cls.parse_obj(d)
 
+    def __hash__(self) -> int:
+        return id(self)
```
nclack, line 209: This is simple and almost always gets the job done but it lacks the property that if `a==b` then `hash(a)==hash(b)`. When things go wrong, it's usually hard to diagnose so I feel like the use of `id()` should be explicitly done by the caller.

Not sure how feasible it is here, but I've used tuples to combine hashable values. Something along the lines of: `hash(tuple(thing.as_ordered_dict().items()))`

tests/test_package_meta.py
```diff
@@ -13,3 +13,7 @@ def test_package_metadata_version():
         description_content_type="text/markdown",
     )
     assert pm3.metadata_version == "2.1"
+
+
+def test_hashable():
+    hash(PackageMetadata(name="test", version="1.0"))
```
nclack, line 19: In the same spirit as the comment above
```suggestion
    assert hash(PackageMetadata(name="test", version="1.0")) == hash(PackageMetadata(name="test", version="1.0"))
    assert hash(PackageMetadata(name="test", version="1.0")) != hash(PackageMetadata(name="test", version="1.1"))
```

nclack (approved):
Left notes around the use of `id()` as a hash.

PR #99, opened 2022-02-10
Fix some typos and dead links

andy-sweet:
I ran into these while working through the plugin docs. I have some other general comments that might lead to other changes, but I though it was worth making a tiny PR to fix these first.

### DIFF
--- a/_docs/templates/_npe2_readers_guide.md.jinja
+++ b/_docs/templates/_npe2_readers_guide.md.jinja
@@ -48,7 +48,7 @@ This demonstrates the now-deprecated `napari-plugin-engine` pattern.
 
 **python implementation**
 
-[hook specificiation](https://napari.org/plugins/stable/hook_specifications.html#napari.plugins.hook_specifications.napari_get_reader)
+[hook specification](https://napari.org/plugins/stable/npe1.html#napari.plugins.hook_specifications.napari_get_reader)
 
 ```python
 from napari_plugin_engine import napari_hook_implementation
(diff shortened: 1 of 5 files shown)


_docs/templates/_npe2_readers_guide.md.jinja
```diff
@@ -57,7 +57,7 @@ from napari_plugin_engine import napari_hook_implementation
 @napari_hook_implementation
 def napari_get_reader(path: PathOrPaths) -> Optional[ReaderFunction]:
     # If we recognize the format, we return the actual reader function
-    if isinstance(path, str) and path.endswith("*.xyz"):
+    if isinstance(path, str) and path.endswith(".xyz"):
```
andy-sweet, line 60: `str.endswith` just does plain string matching not glob-like matching, so using something like this in a plugin doesn't work (unless the filename actually contains an `*` before the `.xyz` I guess?).

_docs/templates/_npe2_writers_guide.md.jinja
```diff
@@ -31,7 +31,7 @@ by the corresponding writer contribution in the manifest).
    **attributes**, (e.g. `{'name': 'My Layer', 'opacity': 0.6}`)
 
    ```python
-   def single_layer_writer(path: str, data: Any, attributes: dict) -> Optional[str]:
+   def single_layer_writer(path: str, data: Any, attributes: dict) -> List[str]:
```
andy-sweet, line 34: Based on the formal type definition below, it seems like this should be `List[str]` instead of `Optional[str]`. In practice, I tried returning `None`, `path`, and `[path]` and they all worked, so may not be too important.

nclack, line 34: I have a hard time remembering where we ended up on this. It looks like types.py also shows `SingleWriterFunction` returning a `List`. So I must be wrong. Looks like the change on 34 is fine.

nclack (approved):
Thanks!

PR #101, opened 2022-02-13
Add enable disable

tlambert03:
This adds:
- `PluginManager.enable`
- `PluginManager.disable`

and new signals:
- `PluginManager.events.plugins_registered`
- `PluginManager.events.activation_changed`
- `PluginManager.events.enablement_changed`

`_ContributionsIndex` has been minimized, but not completely removed, and gains a `remove_contributions` method

**ALSO**: note that this removes `discover()` from the init method.  So it must be called manually

see also https://github.com/napari/napari/pull/4086

I'm also taking the opportunity to re-order the methods and add docstrings on PluginManager (so the number of lines in this PR looks higher than it really is).  They're starting to get scattered about without much logic

### DIFF
--- a/npe2/_plugin_manager.py
+++ b/npe2/_plugin_manager.py
@@ -32,7 +32,6 @@ from .types import PathLike, PythonName
 
 if TYPE_CHECKING:
     from .manifest.commands import CommandContribution
-    from .manifest.contributions import ContributionPoints
     from .manifest.menus import MenuItem
     from .manifest.readers import ReaderContribution
     from .manifest.sample_data import SampleDataContribution
@@ -54,6 +53,9 @@ if TYPE_CHECKING:
         def __getitem__(self, index: Union[int, slice]) -> Set[Interval[T]]:
             ...
 
+        def __iter__(self) -> Iterator[Interval[T]]:
+            ...
+
 
 __all__ = ["PluginContext", "PluginManager"]
 PluginName = str  # this is `PluginManifest.name`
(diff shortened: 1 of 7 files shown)


tlambert03:
this is all set @nclack

npe2/_plugin_manager.py
```diff
@@ -52,105 +54,220 @@ def addi(self, begin: int, end: int, data: T) -> None:
...
+        "whenever new plugins are registered. 'Registered' means that a "
+        "manifest has been provided or discovered.",
+    )
+    activation_changed = Signal(
+        set,
+        set,
+        description="Emitted with two arguments: a set names of plugins "
+        "(strings) that were activated, and a set of names that were "
+        "deactivate. 'Activated' means the plugin has been *imported*, it's "
+        "`on_activate` function called.",
```
nclack, line 194:
```suggestion
        description="Emitted with two arguments: a set of plugin "
        "names that were activated, and a set of names that were "
        "deactivated. 'Activated' means the plugin has been *imported*, it's "
        "`on_activate` function was called.",
```

npe2/_plugin_manager.py
```diff
@@ -52,105 +54,220 @@ def addi(self, begin: int, end: int, data: T) -> None:
...
+    )
+    enablement_changed = Signal(
+        set,
+        set,
+        description="Emitted with two arguments: a set names of plugins "
+        "(strings) that were enabled, and a set of names that were "
+        "disabled. 'Disabled' means the plugin remains installed, but it "
+        "cannot be activated, and its contributions will not be indexed.",
```
nclack, line 202:
```suggestion
        description="Emitted with two arguments: a set of plugin names "
        "that were enabled, and a set of names that were "
        "disabled. 'Disabled' means the plugin remains installed, but it "
        "cannot be activated, and its contributions will not be indexed.",
```

nclack (approved):
Looks good. Made a few small suggestions.

npe2/_plugin_manager.py
```diff
@@ -187,105 +305,132 @@ def activate(self, key: PluginName) -> PluginContext:
...
+
+    def disable(self, plugin_name: PluginName) -> None:
+        """Disable a plugin"""
+        self._disabled_plugins.add(plugin_name)
+        self._contrib.remove_contributions(plugin_name)
+        self.events.enablement_changed({}, {plugin_name})
+
+    def is_disabled(self, plugin_name: str) -> bool:
+        """Return `True` if plugin_name is disabled."""
+        return plugin_name in self._disabled_plugins
```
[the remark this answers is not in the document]
tlambert03: I considered that, but that leads to the confusing possibility of `is_enabled('nonsense') == True`  and more often then not, in this case, we're checking to see `is_disabled` rather than `not is_enabled()`

npe2/_plugin_manager.py
```diff
@@ -52,105 +54,220 @@ def addi(self, begin: int, end: int, data: T) -> None:
...
             if reader.accepts_directories:
                 self._readers[""].append(reader)
-        for writer in ctrb.writers or []:
-            self._writers_by_command[writer.command].append(writer)
+        for writer in ctrb.writers or ():
             for c in writer.layer_type_constraints():
-                self._writers_by_type[c.layer_type].addi(*c.bounds, writer)
+                self._writers[c.layer_type].addi(*c.bounds, writer)
+
+    def remove_contributions(self, key: PluginName) -> None:
```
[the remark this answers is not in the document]
tlambert03: i removed everything that was simple (widgets, samples, themes) but kept these more complicated ones where index performance would actually matter. and, it was easy enough to prune these (whereas widgets and samples didn't keep track of who they belonged to)

npe2/_plugin_manager.py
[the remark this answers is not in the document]
nclack: good point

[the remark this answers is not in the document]
Carreau: Ha, I was coming to say this. 
I mean we still can have `is_disabled('nonsens') == False`. So maybe we should error or check that the `plugin_name` is a known plugin ?

npe2/_plugin_manager.py
```diff
@@ -187,105 +305,132 @@ def activate(self, key: PluginName) -> PluginContext:
...
+
+        self._disabled_plugins.remove(plugin_name)
+        mf = self._manifests.get(plugin_name)
+        if mf is not None:
+            self._contrib.index_contributions(mf)
+        self.events.enablement_changed({plugin_name}, {})
+
+    def disable(self, plugin_name: PluginName) -> None:
+        """Disable a plugin"""
+        self._disabled_plugins.add(plugin_name)
```
Carreau, line 346: You may want to to do the same check that a plugin is not already disabled, even if `remove_contributions` is no-op for consistency whether `self.events.enablement_changed` is emited.

npe2/_plugin_manager.py
```diff
@@ -52,105 +54,220 @@ def addi(self, begin: int, end: int, data: T) -> None:
...
+        "whenever new plugins are registered. 'Registered' means that a "
+        "manifest has been provided or discovered.",
+    )
+    activation_changed = Signal(
+        set,
+        set,
+        description="Emitted with two arguments: a set of plugin "
+        "names that were activated, and a set of names that were "
+        "deactivated. 'Activated' means the plugin has been *imported*, it's "
+        "`on_activate` function was called.",
```
Carreau, line 194:
```suggestion
        "deactivated. 'Activated' means the plugin has been *imported*, its "
        "`on_activate` function was called.",
```

I think... I'm unsure. I doubt now.

npe2/_plugin_manager.py
```diff
@@ -52,105 +54,220 @@ def addi(self, begin: int, end: int, data: T) -> None:
...
+        return cls.__instance
 
     @property
     def commands(self) -> CommandRegistry:
         return self._command_registry
 
-    def discover(self, paths: Sequence[str] = ()) -> None:
+    # Discovery, activation, enablement
+
+    def discover(self, paths: Sequence[str] = (), clear=False) -> None:
```
Carreau, line 234: I believe technically the super class also allows Path in addition to strings.

npe2/_plugin_manager.py
```diff
@@ -52,105 +54,220 @@ def addi(self, begin: int, end: int, data: T) -> None:
...
+        clear : bool
+            Clear and re-index the environment.  If `False` (the default),
+            calling discover again will only register and index newly
+            discovered plugins. (Existing manifests will not be re-indexed)
         """
-        self._contrib = _ContributionsIndex()
-        self._manifests.clear()
+        if clear:
+            self._contrib = _ContributionsIndex()
+            self._manifests.clear()
```
Carreau, line 249: it's not evented (I think ?) so  you can also just assign an empty dict ? Would it be faster ? Unless something else keep reference to `_manifest` ?

npe2/_plugin_manager.py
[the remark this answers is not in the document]
tlambert03: That's `__contains__`

npe2/_plugin_manager.py
```diff
@@ -52,105 +54,220 @@ def addi(self, begin: int, end: int, data: T) -> None:
...
+        if key not in self._indexed:
+            return  # pragma: no cover
+
+        for cmd_id, (_, plugin) in list(self._commands.items()):
+            if key == plugin:
+                del self._commands[cmd_id]
+        for contribs in self._readers.values():
+            for i, ctrb in reversed(list(enumerate(contribs))):
+                if ctrb.plugin_name == key:
+                    contribs.pop(i)
```
Carreau, line 102: Aren't we starting to make things really complicated using dicts, instead of using something flatter, and filtering in the relevant methods ?

npe2/_plugin_manager.py
tlambert03: yes,  you're right :joy:  this is one of those things that my brain knows and my fingers almost always do wrong :)

npe2/_plugin_manager.py
tlambert03: perhaps!  not married at all to this one if you want to propose a better structure

npe2/_plugin_manager.py
tlambert03: i'm trying to keep all private access within classes, even within npe2, so it should be fine just to set to empty dict

npe2/_plugin_manager.py
tlambert03: good point.  I think we want to remove discover from PluginManifest altogether and move that logic here.  so might do in a followup pr

npe2/_plugin_manager.py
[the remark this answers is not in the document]
Carreau: No, I think that's fine.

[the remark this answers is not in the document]
tlambert03: wait, fine to leave it?  or fine to have enabled return true for uninstalled plugins? :joy:?

npe2/_plugin_manager.py
Carreau: Ok, let' try to finish that and I'll suggest a refactor.

npe2/_plugin_manager.py
Carreau: ["I heard you're idea's and their definately good"](https://xkcd.com/1238/)

Honestly it might be easier for non-native speakers.

Carreau (approved):

PR #102, opened 2022-02-15
DOC: typo missing backtick

Carreau:


nclack (approved):
thanks! Looks like there's a pre-commit failure to fix.

PR #105, opened 2022-02-16
Clarify Typing.

Carreau:
From what I can tell, those are always path-or-paths from the types.

From the implementation that may vary. I think it's actually never Path,
but always str.

### DIFF
--- a/npe2/io_utils.py
+++ b/npe2/io_utils.py
@@ -38,7 +38,7 @@ def read(path: PathLike, *, plugin_name: Optional[str] = None) -> List[LayerData
 
 
 def read_get_reader(
-    path: PathLike, *, plugin_name: Optional[str] = None
+    path: Union[str, Sequence[str]], *, plugin_name: Optional[str] = None
 ) -> Tuple[List[LayerData], ReaderContribution]:
     """Variant of `read` that also returns the `ReaderContribution` used."""
     return _read(path, plugin_name=plugin_name, return_reader=True)
@@ -91,7 +91,7 @@ def write_get_writer(
 
 @overload
 def _read(
-    path: PathLike,
+    path_or_paths: PathOrPaths,
     *,
     plugin_name: Optional[str] = None,
     return_reader: Literal[False] = False,
@@ -102,7 +102,7 @@ def _read(
 
 @overload
 def _read(
-    path: PathLike,
+    path_or_paths: PathOrPaths,
     *,
     plugin_name: Optional[str] = None,
     return_reader: Literal[True],
@@ -112,7 +112,7 @@ def _read(
 
 
 def _read(
-    path: PathLike,
+    path_or_paths: PathOrPaths,
     *,
     plugin_name: Optional[str] = None,
     return_reader: bool = False,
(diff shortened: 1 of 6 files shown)


tlambert03:
Yeah, I've been wanting to clarify this too.  We _might_ need to double check this on the napari side, to make sure that all path objects have indeed been coerced to string before we call out to npe2.  but this is definitely how I think it should be (no path objects by this point)

Carreau:
How would you feel about not trying to be a bit stricter about what we pass to read ?
I thinks that either one or a list of either strings or path is a 4 item matrix that starts to get in the way of readability and consistency. 

I think that having npe2 always normalising to a list of string would be beneficial. Unless there is a semantic difference between a (list of 1 item) and a single item on it's own, I both believe that it would benefit everyone.

Carreau:
> Yeah, I've been wanting to clarify this too. We _might_ need to double check this on the napari side, to make sure that all path objects have indeed been coerced to string before we call out to npe2. but this is definitely how I think it should be (no path objects by this point)

Ok, then I think we should start putting assert/warnings everywhere.

tlambert03:
ok!  those are real cheap right?

tlambert03:
lol.  fair point :)

"OMGZ ... WHAT ABOUT THE PERFORMANCE IMPACT??"

Carreau:
> WHAT ABOUT THE PERFORMANCE IMPACT

Switch to Julia.

tlambert03:
> And what do you think about always calling the contributions with a list of path ?

unfortunately, that was claimed semantically early on to mean "please stack these paths yourself".  Not that I like that pattern... but we can't just change that without a lot of communication

Carreau:
And how many users current use npe2 readers contributions ? Would the migration to npe2 be a good time to do this ?

tlambert03:
conceivably... but it's a bigger conversation.  I'm more than happy to have it though!  We'd need to add a `stack=True` argument to both the get_reader and the reader i guess, which means we either need to break API or do inspection.  All doable, just some decisions to be made and needs a clear plan

npe2/io_utils.py
```diff
@@ -38,9 +38,10 @@ def read(path: PathLike, *, plugin_name: Optional[str] = None) -> List[LayerData
 
 
 def read_get_reader(
-    path: PathLike, *, plugin_name: Optional[str] = None
+    path: Union[str, Sequence[str]], *, plugin_name: Optional[str] = None
 ) -> Tuple[List[LayerData], ReaderContribution]:
     """Variant of `read` that also returns the `ReaderContribution` used."""
+    _ensure_str_or_seq_str(path)
```
tlambert03, line 44: this (and `read`) is actually one place where I think we should allow Path objects.  This is part of the public api that someone (besides napari) can use to take advantage of napari plugins without having to open up napari.  In other words, these functions are there to satisfy the requests in https://github.com/napari/napari-plugin-engine/issues/24

Carreau: I'm happy to cast here, though from a [github search](https://github.com/search?q=read_get_reader&type=code) it's only used on a test Repo of Juan, so I _think_ we still have the opportunity to tell users that they should pass in strings.

tlambert03 (approved):
this looks good to me @Carreau.  is it ready to go?  I can go either way on the side conversation about the public `npe2.read` signature

PR #107, opened 2022-02-19
Simplify Reader/writer internal logic.

Carreau:
Sorry, in a train, tethering, a couple of comments inline. 

This is better read as 3 independant commits with each descriptions inlined below. 

Will come back to it and edit, but wanted to push it here to give you something to read :-) 
---- 

Simplify reader manipulation.

Migrate from a dict of list with complex mutations, with a flatter table
that we completely filter everytime.

Even with a few thousand readers this should still be quite fast and
simplify the code a bit. (Boring is good). We also make the reader
hashable even if technically the filename patterns are mutable (maybe
they should not be?) to simplify manipulation a bit more.

----  

Simplify writers: remove interval tree.

Trying to manipulate interval-trees is IMHO more complex that what we
need to, and instead of having a dict-of-interval tree, it is almost as
simple to store a table and filter the table on removal.

It is a bit less efficient optimally in some cases, but that should
still be sufficiently fast even with thousands of writers/layer
constraints.

It also remove a dependency, which in this case is I believe a plus.
We could still go 1/2 way and have writers be a dict of (min,max,
writers) but we also don't have _that many_ layer types so it's not
going to be much more performant.

In the end if you squint this really look like sql, as we select writers
based on the layer types,bounds and order-by based on some categories,
So yes this is not sql as we have live objects, and q sql-like library
would be overkill... for now.

----

Simplify iteration a bit.

We just start with all the writers and trim down.
Really in the end we are reinventing a sort-of sql.

### DIFF
--- a/_docs/render.py
+++ b/_docs/render.py
@@ -183,6 +183,8 @@ def main(dest: Path = _BUILD):
 
     dest.mkdir(exist_ok=True, parents=True)
     schema = PluginManifest.schema()
+    # this does not work offline.
+    # it should also likely not rely on latest release, should it ?
     with urlopen(SCHEMA_URL) as response:
         schema = json.load(response)
 
--- a/npe2/_plugin_manager.py
+++ b/npe2/_plugin_manager.py
@@ -66,7 +66,7 @@ class _ContributionsIndex:
     def __init__(self) -> None:
         self._indexed: Set[str] = set()
         self._commands: Dict[str, Tuple[CommandContribution, PluginName]] = {}
-        self._readers: DefaultDict[str, List[ReaderContribution]] = DefaultDict(list)
+        self._readers: List[Tuple(str, ReaderContribution)] = list()
         self._writers: DefaultDict[
             LayerType, TypedIntervalTree[WriterContribution]
         ] = DefaultDict(IntervalTree)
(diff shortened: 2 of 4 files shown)


nclack:
> writers: remove interval tree

I could go either way on this one. It's definitely nice to remove a dependency. Agree that performance isn't important. I find the code easier to understand using intervaltree, but then again I wrote it :)

Carreau:
> I find the code easier to understand using intervaltree, but then again I wrote it :)

THe writer part was not too hard to understand once you get what the interval tree was doing, though it was a bit tough to modify as you have to be aware of the different methods like like `addi(...)`, and the use of `.data`.

I made it a separate commit if we don't want to remove it. My main concern was 
```
        for contribs in self._readers.values():
            for i, ctrb in reversed(list(enumerate(contribs))):
                if ctrb.plugin_name == key:
                    contribs.pop(i)
```

which was was a codesmell for me. Having to iterate in reverse order to correctly mutate is a codesmell to me.

tlambert03:
> Having to iterate in reverse order to correctly mutate is a codesmell to me.

happy to change that, what pattern do you like to use to remove certain items from a list when they match a condition?

tlambert03:
all the suggestions in the original post sound fine to me.  also agnostic on the interval tree, happy to go with whatever you two decide

Carreau:
> happy to change that, what pattern do you like to use to remove certain items from a list when they match a condition? (i know one can create a new list and re-assign... didn't dig deeper to see if that would be ok here)

Here ai changed the default dict of list to a flat list, and filter/reassing each time.

nclack (approved):
looks good to me.

_docs/render.py
```diff
@@ -183,6 +183,8 @@ def main(dest: Path = _BUILD):
 
     dest.mkdir(exist_ok=True, parents=True)
     schema = PluginManifest.schema()
+    # this does not work offline.
+    # it should also likely not rely on latest release, should it ?
```
tlambert03, line 187:
```suggestion
```

(unrelated to this PR)

tlambert03 (approved):
if @nclack is happy with the removal of interval tree, it's fine with me!

thanks @Carreau

Carreau:
Thanks ! And sorry for having missed last plugin meeting.

PR #109, opened 2022-02-21
Push stack=... through all the reader internal API

Carreau:
There is a tiny bit more we can do once
https://github.com/napari/napari/pull/4130 is merged and we assume a
minimal version of napari that only passes list of str to npe2 and give
it the stack=parameter as well, but it will take a couple releases
for this to be the case


tlambert03:
napari 4130 is merged.  this needs a rebase though.  Can you let me know when it's ready for final review?

nclack:
closing in favor of #142

Issue #108, opened 2022-02-20
pytest tests/test_docs.py needs internet connection.


Carreau:
* npe2 version:
* Python version:
* Operating System:

### Description

`test_render_docs` imports main from `from _docs.render`, whcih itself uses urlopen with  "https://github.com/napari/npe2/releases/latest/download/schema.json" as a parameter.

I think this might be a problem in other cases, like packaging on linux, or other system, and in general we should not get the schema from latest napari from the test suite. 

I would bundle the schema and have a separate github action step that check the schema is up to date, but not rely in having an internat  connection during the tests.

### What I Did

```
$ cd /mnt/train_with_no_wifi
$ pytest
```

Test suite fails.

PR #111, opened 2022-02-21
Store schema locally

Carreau:
See #108

### DIFF
--- a/.github/workflows/sync_schema.yml
+++ b/.github/workflows/sync_schema.yml
@@ -1,0 +1,27 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "v*"
+  pull_request: {}
+  workflow_dispatch:
+
+jobs:
+  test:
+    name: Test Schema Sync
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: []
+
+    steps:
+      - uses: actions/checkout@v2
+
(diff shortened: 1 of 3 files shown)


.github/workflows/sync_schema.yml
```diff
@@ -0,0 +1,27 @@
+name: CI
```
Carreau, line 1:
```suggestion
name: Schema Sync
```

tlambert03:
Curious what @nclack and @czaki think here. 

#83 was going to do this and then specifically closed after discussion in meeting

I could be swayed, but just wanna hear thoughts again

tlambert03:
Minimally, @Carreau, I think we'll  need to be careful when checking out the schema elsewhere (eg for docs), to specifically checkout the latest release, and not main.

Czaki:
But why store schema locally instead of generating it dynamically?

nclack:
The main problem here is that `test_render_docs` requires an internet connection. I think it's reasonable to fall back to using the local schema for the purposes of testing when urlopen fails.  I think this just needs a try around the `urlopen`-block in `render.py` (see #112).

To me `schema.json` feels like a kind of documentation artifact so I'd rather not commit it back to the repo.

tlambert03:
Yeah; since we've already been through this discussion and closed a related PR to this effect, I'm going to close this.  We can discuss in group meeting anything we may have overlooked.  I agree with falling back to rebuilding the schema if there's no Internet.

PR #112, opened 2022-02-26
doc gen: fall back to local schema when urlopen fails

### LINKED ISSUE #108: pytest tests/test_docs.py needs internet connection.

nclack:
Closes #108 

See also discussion in #111. This is an alternate approach.

### DIFF
--- a/_docs/render.py
+++ b/_docs/render.py
@@ -10,6 +10,7 @@ from inspect import getsource
 from pathlib import Path
 from types import FunctionType
 from typing import Dict, Optional, Set
+from urllib.error import URLError
 from urllib.request import urlopen
 
 import yaml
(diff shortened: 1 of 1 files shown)


_docs/render.py
```diff
@@ -183,8 +184,11 @@ def main(dest: Path = _BUILD):
 
     dest.mkdir(exist_ok=True, parents=True)
     schema = PluginManifest.schema()
-    with urlopen(SCHEMA_URL) as response:
-        schema = json.load(response)
+    try:
+        with urlopen(SCHEMA_URL) as response:
+            schema = json.load(response)
+    except URLError:
+        print(f"Using local schema as a fallback. Failed to retrieve: {SCHEMA_URL}")
```
Czaki, line 191:
```suggestion
        print(f"Using local schema as a fallback. Failed to retrieve: {SCHEMA_URL}", file=sys.stderr)
```

nclack:
closing in favor of #114

PR #114, opened 2022-03-04
Separate dev test from integration test.

### LINKED ISSUE #108: pytest tests/test_docs.py needs internet connection.

Carreau:
Add a marker that is not ran by default, and add a CI entry to run only
this test

Alternate to #112, #111, closes #108

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,8 @@ jobs:
           python -m pip install --upgrade pip
           pip install -e .[docs,testing]
 
+      - name: Test Main docs build
+        run: pytest --color yes -m github_main_only
       - name: Test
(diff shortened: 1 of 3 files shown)


nclack (approved):
looks good to me!

tlambert03:
Lovely

PR #116, opened 2022-03-09
add back deprecated _samples on contributions index

tlambert03:
Looks like I unfortunately [accessed the private `plugin_manager._contrib` object in the npe2 module](https://github.com/napari/napari/blob/4c65afcab0e64d90284a1279f3b707228bb4606d/napari/plugins/_npe2.py#L255) in napari (yet another reminder that even code authors shouldn't use private variables across packages! 😂 ).

#101 modified it a bit, which means that if we were to release a new version of npe2 (which we need to for enable disable), then it would break new napari installations of v0.4.14 and v0.4.15. 

This PR adds back the line needed to maintain compatibility with older versions.
(tested on napari v0.4.14 and napari main)

### DIFF
--- a/npe2/_plugin_manager.py
+++ b/npe2/_plugin_manager.py
@@ -9,6 +9,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
+    DefaultDict,
     Dict,
     Iterable,
     Iterator,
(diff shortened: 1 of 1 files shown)


nclack (approved):
thanks!

PR #120, opened 2022-03-11
Fix auto-discovery of plugins for napari <= 0.4.15

tlambert03:
In #101 greedy discovery of plugins was removed from the plugin manager `__init__`, which means that the user of the plugin manager (napari) should call it when it wants.  But napari <=0.4.15 didn't do that.  so this patch checks the napari version and does it for backwards compatibility.  

I yanked the 0.2.0 release and will release this as 0.2.1 today

Note, this also extends the CI napari tests to include `napari/layers` and `napari/components` as those are actually the tests that failed without this (since they were expecting napari-svg to be discovered... but it wasnt)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
           python -m pip install -e ./napari-from-github
           python -m pip install -e .
           # bare minimum required to test napari/plugins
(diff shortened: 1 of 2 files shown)


tlambert03:
remaining failing napari test is unrelated to this (it's a headless test issue fixed in https://github.com/napari/napari/pull/4245)

PR #123, opened 2022-03-12
drop py3.7 & misc small reformats

tlambert03:
no hurry on this one. but now that napari doesn't support 3.7, we can drop it here too

### DIFF
--- a/npe2/manifest/commands.py
+++ b/npe2/manifest/commands.py
@@ -92,9 +92,11 @@ class CommandContribution(BaseModel):
     def exec(
         self,
         args: tuple = (),
-        kwargs: dict = {},
+        kwargs: dict = None,
         _registry: Optional["CommandRegistry"] = None,
     ) -> Any:
+        if kwargs is None:
+            kwargs = {}
         if _registry is None:
             from .._plugin_manager import PluginManager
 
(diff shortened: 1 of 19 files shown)


npe2/manifest/commands.py
```diff
@@ -92,9 +92,11 @@ class Config:
     def exec(
         self,
         args: tuple = (),
-        kwargs: dict = {},
+        kwargs: dict = None,
         _registry: Optional["CommandRegistry"] = None,
     ) -> Any:
+        if kwargs is None:
+            kwargs = {}
```
nclack, line 99: The `kwargs: dict={}` pattern better seems better to me?

pyproject.toml
```diff
@@ -8,3 +8,7 @@ write_to = "npe2/_version.py"
 
 [tool.check-manifest]
 ignore = ["npe2/_version.py", ".pre-commit-config.yaml", ".github_changelog_generator"]
+
+[tool.black]
+target-version = ['py38', 'py39', 'py310']
+line-length=88
```
nclack, line 14: bold move :)

nclack (approved):
nice to have. thanks

npe2/manifest/commands.py
tlambert03: that was a suggested refactor to avoid [mutable defaults](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).  but in this case it's probably fine

PR #124, opened 2022-03-13
NPE1Shim Part 1 - updated _from_npe1 conversion logic to prepare for locally defined objects

tlambert03:
Breaking up #86 into three steps.  This is **part 1**

- A lot of changes have been made to the`_from_npe1` module itself to prepare for the shim logic of being able to recover a _locally_ defined objects.
- A more realistic `npe1-plugin`, that contains a number of tricky scenarios with locally defined widgets and sample data generators.
- The general idea is that if an npe1 plugin creates a locally defined object within the scope of a hook implementation, then we store a shim "python_name" in the form `__npe1shim__.hook_module::hoo_impl_shimidx`.  For example: `__npe1shim__.npe1_module:napari_experimental_provide_dock_widget_2` means "call the `napari_experimental_provide_dock_widget` function in the `npe1_module` and retrieve the 2nd object in the list
- dependence on `napari-plugin-engine` has been completely removed, since all we really needed was the `iter_hookimpls` logic which has been recreated here.

note, steps 1 and 2 have 100% test coverage, but this one doesn't, as it really needs logic from those later PRs to cover it.

### DIFF
--- a/npe2/_from_npe1.py
+++ b/npe2/_from_npe1.py
@@ -1,18 +1,20 @@
 import ast
-import itertools
+import inspect
 import re
 import sys
 import warnings
 from configparser import ConfigParser
 from dataclasses import dataclass
 from functools import lru_cache
+from importlib import import_module
 from pathlib import Path
+from types import ModuleType
 from typing import (
     Any,
     Callable,
     DefaultDict,
     Dict,
-    Iterable,
+    Iterator,
     List,
     Optional,
     Tuple,
@@ -20,17 +22,12 @@ from typing import (
     cast,
 )
 
-from napari_plugin_engine import (
-    HookCaller,
-    HookImplementation,
-    PluginManager,
-    napari_hook_specification,
-)
-
 from npe2.manifest import PluginManifest
 from npe2.manifest.commands import CommandContribution
 from npe2.manifest.themes import ThemeColors
+from npe2.manifest.utils import SHIM_NAME_PREFIX, import_python_name, merge_manifests
 from npe2.manifest.widgets import WidgetContribution
+from npe2.types import WidgetCreator
 
 try:
     from importlib import metadata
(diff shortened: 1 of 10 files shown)


tests/conftest.py
```diff
@@ -1,11 +1,17 @@
...
 
 import pytest
 
 from npe2 import PluginManager, PluginManifest
 
+try:
+    from importlib import metadata
+except ImportError:
+    import importlib_metadata as metadata  # type: ignore
+
```
nclack, line 14: Are we 3.8+ now? If so is this still necessary?

npe2/_from_npe1.py
```diff
@@ -403,8 +426,79 @@ def _safe_key(key: str) -> str:
...
+        obj_name = getattr(obj, "__qualname__", "")
+        if obj_name and "<locals>" not in obj_name:
+            mod = inspect.getmodule(obj) or inspect.getmodule(hook)
+            if mod:
+                mod_name = mod.__name__
+
+    if not (mod_name and obj_name) and (hook and shim_idx is not None):
+        # we weren't able to resolve an absolute name... if we are shimming, then we
+        # can create a special py_name of the form `__npe1shim__.hookfunction_idx`
+        return f"{SHIM_NAME_PREFIX}{_python_name(hook)}_{shim_idx}"
```
nclack, line 492: maybe this gets hit by later pr's but I'm not seeing test coverage here. This means in turn that `_utils._import_npe1_shim` doesn't get called during tests.  Mostly just noting this here so I'll remember to double check later.

nclack (approved):
lgtm.  It was very helpful talking about this earlier.  I made some notes where I wasn't seeing test coverage, but I think the later pr's get those.  Those notes are mostly for me.  I should get to the other pr's later tonight.

tlambert03:
Yeah correct.  Mentioned this in the original post, but this one can merge without 100% coverage, tests are in part 2

npe2/manifest/utils.py
```diff
@@ -153,16 +159,131 @@ def __str__(self) -> str:
...
-    match = PYTHON_NAME_PATTERN.match(python_name)
-    if not match:  # pragma: no cover
-        raise ValueError(f"Invalid python name: {python_name}")
+    if python_name.startswith(SHIM_NAME_PREFIX):
+        return _import_npe1_shim(python_name)
 
-    module_name, funcname = match.groups()
+    _validators.python_name(python_name)  # shows the best error message
+    match = _validators.PYTHON_NAME_PATTERN.match(python_name)
+    module_name, funcname = match.groups()  # type: ignore [union-attr]
```
Carreau, line 227: just split on `:` ? I guess using regex to validate even more sure, but if it's invalid it will anyway fail just after.

npe2/_from_npe1.py
```diff
@@ -39,28 +38,42 @@
...
+        self,
+        function: Callable,
+        plugin: Optional[ModuleType] = None,
+        plugin_name: Optional[str] = None,
+        **kwargs,
+    ):
+        self.function = function
+        self.plugin = plugin
+        self.plugin_name = plugin_name
+        self._specname = kwargs.get("specname")
```
Carreau, line 55: Hum, why not:
```suggestion
        *,
        specname=None
    ):
        self.function = function
        self.plugin = plugin
        self.plugin_name = plugin_name
        self._specname = specname
```

Are you afraid people will see specname in the signature ? 
(we can also keep `**kwargs`, if you want to allow passing more arguments. )

npe2/_from_npe1.py
```diff
@@ -98,102 +106,112 @@ def plugin_packages() -> List[PluginPackage]:
...
     Parameters
     ----------
-    plugin_name : str
-        Name of package/plugin to convert, by default None
-    module : Module
+    plugin : Union[str, metadata.Distribution, None]
+        Name of package/plugin to convert.  Or a `metadata.Distribution` object.
+        If a string, this function should be prepared to accept both the name of the
+        package, and the name of an npe1 `napari.plugin` entry_point. by default None
+    module : Optional[Module]
```
Carreau, line 124: Type in docstring does not match type in signature, maybe we can go "ModuleType" 

We can always change later to not mess up the 3 PR chains.

Carreau:
+1, I can't merge because of codecov, should I deactivate it in setting for now ? It is likely due for github issues yesterday.

Carreau:
pushed an empty commit trying to kick CI to get coverage and be able to merge

PR #125, opened 2022-03-13
NPE1Adapter Part 2 - adding the NPE1Adapter object.

tlambert03:
Breaking up https://github.com/napari/npe2/pull/86 into three steps. This is **part 2** (but it contains #124 which can be reviewed first)

This adds the `NPE1Adapter` itself, and makes the plugin manager aware of it.
 
- in the discovery process, the `_from_dist` function now creates either a `PluginManifest` or an `NPE1Shim` based on the entry points offered by the plugin
- `NPE1Adapter` is a subclass of `PluginManifest` that can lazily load contributions (when `manifest.contributions` is accessed), by calling `manifest_from_npe1` which was updated in #124
- This lazily loaded manifest is used for the duration of the session, (caching behavior will be added in step 3)
- The plugin manager itself now saves a list of `_npe1_adapters` during plugin discovery and gains a new method `index_npe1_adapters` that napari can use to force importing (or loading from cache) of all npe1 plugins.

Caching behavior comes in the next PR

### DIFF
--- a/npe2/_plugin_manager.py
+++ b/npe2/_plugin_manager.py
@@ -59,6 +59,10 @@ class _ContributionsIndex:
             list
         )
 
+    def reindex(self, manifest):
+        self.remove_contributions(manifest.name)
+        self.index_contributions(manifest)
+
     def index_contributions(self, manifest: PluginManifest):
         ctrb = manifest.contributions
         if not ctrb or manifest.name in self._indexed:
@@ -298,6 +302,13 @@ class PluginManager:
             self._contrib.index_contributions(manifest)
         self.events.plugins_registered.emit({manifest})
 
+    def unregister(self, key: PluginName):
+        if key not in self._manifests:
+            raise ValueError(f"No registered plugin named {key!r}")  # pragma: no cover
+        self.deactivate(key)
+        self._contrib.remove_contributions(key)
+        self._manifests.pop(key)
+
     def activate(self, key: PluginName) -> PluginContext:
         """Activate plugin with `key`.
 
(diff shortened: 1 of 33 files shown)


npe2/_plugin_manager.py
```diff
@@ -220,6 +221,7 @@ def __init__(
         self._contrib = _ContributionsIndex()
         self._manifests: Dict[PluginName, PluginManifest] = {}
         self.events = PluginManagerEvents(self)
+        self._shims: List[NPE1Shim] = []
```
nclack, line 224: Maybe rename this field to `_npe1_shims`.  Similarly, `test_shims.py` -> `test_npe1_shims.py`. I feel like being a bit pedantic with names will keep it clear what this was for later on. I think this and the test are the only spots where I felt a rename would be nice. The rest is clear.

npe2/manifest/_npe1_shim.py
```diff
@@ -0,0 +1,45 @@
...
+try:
+    from importlib import metadata
+except ImportError:
+    import importlib_metadata as metadata  # type: ignore
+
+
+logger = logging.getLogger(__name__)
+
+
+class NPE1Shim(PluginManifest):
```
nclack, line 17: A comment introducing what the general motivation and approach behind the shim is might go here (or maybe module level).

nclack (approved):
looks good. The docs in _from_npe1 are good.  I thought a comment giving a bit of a high level introduction to what the shim was could go in _npe1_shim.py.

npe2/manifest/_npe1_shim.py
tlambert03: also, describe how the general inspection works and document it somewhere clearly

npe2/_plugin_manager.py
tlambert03: renamed everything to `NPE1Adapter`  `_npe1_adapters`, etc...

PR #126, opened 2022-03-13
NPE1Adapter Part 3 - caching of adapter manifests

tlambert03:
Breaking up https://github.com/napari/npe2/pull/86 into three steps. This is **part 3** (but it contains #124 and #125 which can be reviewed first)

This PR adds local caching of adapter manifests after contributions have been imported and indexed.  This is the thing that will really make napari much faster on load time.

- This lazily loaded manifest is used for the duration of the session, and also cached for use in the next session
- caching will _not_ be done for packages that are installed in editable mode.
- the `npe2 cache` command provides additional ways to query/modify the cache 
- cache can be cleared by using `npe2 cache --clear`.
- Cache can be ignored per session by starting napari with `NPE2_NOCACHE=1`

### DIFF
--- a/npe2/manifest/_npe1_shim.py
+++ b/npe2/manifest/_npe1_shim.py
@@ -1,4 +1,11 @@
 import logging
+import os
+import site
+from pathlib import Path
+from shutil import rmtree
+from typing import List, Sequence
+
+from appdirs import user_cache_dir
 
 from .._from_npe1 import manifest_from_npe1
 from .package_metadata import PackageMetadata
@@ -11,6 +18,34 @@ except ImportError:
 
 
 logger = logging.getLogger(__name__)
+SHIM_CACHE = Path(user_cache_dir("napari", "napari")) / "npe2" / "shims"
+NPE2_NOCACHE = "NPE2_NOCACHE"
+
+
+def clear_cache(names: Sequence[str] = ()) -> List[Path]:
+    """Clear cached npe1 shim files
+
+    Parameters
+    ----------
+    names : Sequence[str], optional
+        selection of plugin names to clear, by default, all will be cleared
+
+    Returns
+    -------
+    List[Path]
+        List of filepaths cleared
+    """
+    _cleared: List[Path] = []
+    if SHIM_CACHE.exists():
+        if names:
+            for f in SHIM_CACHE.glob("*.yaml"):
+                if any(f.name.startswith(f"{n}_") for n in names):
+                    f.unlink()
+                    _cleared.append(f)
+        else:
+            _cleared = list(SHIM_CACHE.iterdir())
+            rmtree(SHIM_CACHE)
+    return _cleared
 
 
 class NPE1Shim(PluginManifest):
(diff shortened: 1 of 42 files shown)


npe2/manifest/_npe1_shim.py
```diff
@@ -0,0 +1,113 @@
...
+            self.contributions = mf.contributions
+            logger.debug("%r npe1 shim loaded from cache", self.name)
+            return
+
+        with discovery_blocked():
+            try:
+                mf = manifest_from_npe1(self._dist, shim=True)
+            except Exception as e:
+                warnings.warn(
+                    f"Failed to detect contributions for np1e plugin {self.name!r}: {e}"
```
nclack, line 81:
```suggestion
                    f"Failed to detect contributions for npe1 plugin {self.name!r}: {e}"
```

npe2/manifest/_npe1_shim.py
```diff
@@ -0,0 +1,113 @@
...
+        return super().__getattribute__(__name)
+
+    def _load_contributions(self) -> None:
+        """imports and inspects package using npe1 plugin manager"""
+
+        self._is_loaded = True  # if we fail once, we still don't try again.
+        if self._cache_path().exists() and not os.getenv(NPE2_NOCACHE):
+            mf = PluginManifest.from_file(self._cache_path())
+            self.contributions = mf.contributions
+            logger.debug("%r npe1 shim loaded from cache", self.name)
```
nclack, line 73: First, I love the use of logger.debug.  Yay for logging.

Second, I think it's fine for internal use but I'm a bit worried that the term `npe1` is something devs will understand but users/new-folks won't? not sure if that matters for a debug message.

nclack:
Testing this out I ran into something I don't understand.  This is on a py39 env with napari main, this pr and napari_animation (which is npe1) installed.  If I run python and import napari_animation I get:

```
% python
>>> import napari_animation
/Users/nclack/src/CZI/npe2/npe2/manifest/_npe1_shim.py:80: UserWarning: Failed to detect contributions for npe1 plugin 'napari-tools-menu': cannot import name 'get_settings' from partially initialized module 'napari.settings' (most likely due to a circular import) (/Users/nclack/src/CZI/napari/napari/settings/__init__.py)
  warnings.warn(
```

which seems strange? I'm out of time tonight, but I'll try to reproduce in the morning with a clean env. 

On the success side, I can see e.g. napari-ndtiffs get cached.  I really like that there's cli support.

nclack (commented):
I ran into an issue and got confused. Will pick this up in the am.

tlambert03:
I'll take a look soon. I'll remind you that napari tools menu is not really a plugin, but rather a monkey patch that doesn't have hook implementations, so all bets are off for that one .

npe2/manifest/utils.py
```diff
@@ -153,16 +159,131 @@ def __str__(self) -> str:
...
+        The <index>th object returned from the callable <hook_python_name>.
+
+    Raises
+    ------
+    IndexError
+        If len(<hook_python_name>()) <= <index>
+    """
+
+    assert shim_name.startswith(SHIM_NAME_PREFIX), f"Invalid shim name: {shim_name}"
+    python_name, idx = shim_name[13:].rsplit("_", maxsplit=1)  # TODO, make a function
```
nclack, line 191: I'm not quite sure what the TODO is referring to. I'm not sure I would factor anything out of this function as is.  Should probably avoid the `13` - replace with `len(SHIM_NAME_PREFIX)` etc.

nclack (approved):
Looks good. There was one spot where you hardcode the length of SHIM_NAME_PREFIX and I left a couple of questions.

One question I had was how to document some of the inferences that get made when converting plugins (like categorizing a theme as dark/light mode based on luma).  Or extracting glob patterns for readers.  I feel like that's a future discussion though.

npe2/manifest/_npe1_shim.py
tlambert03: left for debug... but changed everywhere else to "first generation napari plugin"

npe2/manifest/_npe1_shim.py
```diff
@@ -0,0 +1,113 @@
...
+from .schema import PluginManifest, discovery_blocked
+
+try:
+    from importlib import metadata
+except ImportError:
+    import importlib_metadata as metadata  # type: ignore
+
+
+logger = logging.getLogger(__name__)
+SHIM_CACHE = Path(user_cache_dir("napari", "napari")) / "npe2" / "shims"
```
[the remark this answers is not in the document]
tlambert03: generally changed to "adapter", so this is `ADAPTER_CACHE`

tlambert03:
> /Users/nclack/src/CZI/npe2/npe2/manifest/_npe1_shim.py:80: UserWarning: Failed to detect contributions for npe1 plugin 'napari-tools-menu': cannot import name 'get_settings' from partially initialized module 'napari.settings' (most likely due to a circular import) 

wasn't able to repeat in a clean environment with napari, npe2, and napari-animation... but if I `pip install napari-tools-menu` i do get it.  will dig

tlambert03:
> wasn't able to repeat in a clean environment with napari, npe2, and napari-animation... but if I `pip install napari-tools-menu` i do get it. will dig

yeah, this circular import is because napari-tools-menu imports the private `napari._qt` module [here](https://github.com/haesleinhuepf/napari-tools-menu/blob/main/napari_tools_menu/__init__.py#L9).  sorry, that's a "will not fix" :joy:

npe2/manifest/_npe1_shim.py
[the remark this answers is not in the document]
tlambert03: i think you nailed it with your word "adapt" ... that's the new word :)

tlambert03:
merging!  

This by no means implies that we can't iterate / refactor / cleanup moving forward.  but it will make it a bit easier to get this in and play with it a bit

PR #128, opened 2022-03-13
Add DynamicPlugin object/context for ease of testing & dynamic plugin creation

tlambert03:
was starting to work on some stuff over in https://github.com/napari/napari/pull/4015 and realized we need a better way to inject temporary functionality into a test plugin manager.  This adds a `DynamicPlugin` context manager that controls a manifest, and allows you to register commands with decorators:

```py
from npe2 import DynamicPlugin

with DynamicPlugin() as plugin:
    @plugin.contribute.sample_data
    def make_image(x):
        ...
```

it will handle creation of a command, and the corresponding contribution type, and inject it into the plugin manifest and reindex.  The decorator takes all the same parameters that the corresponding contribution would take:

```python
    @plugin.contribute.writer(layer_types=["image"])
    def write_path(path, layer_data):
        ...
```

it cleans up after itself when leaving the context

**edit**: this also modifies a small amount of logic on the unregistering/deactivation of plugins in general, specifically around unregistering their commands


npe2/_command_registry.py
```diff
@@ -96,6 +99,20 @@ def unregister(self, id: str):
...
 
+    def register_manifest(self, mf: PluginManifest) -> None:
+        """Register all commands in a manifest"""
+        if mf.contributions and mf.contributions.commands:
+            for cmd in mf.contributions.commands:
+                if cmd.python_name and cmd.id not in self:
+                    self.register(cmd.id, cmd.python_name)
+
+    def unregister_manifest(self, mf: PluginManifest) -> None:
+        """Register all commands in a manifest"""
```
nclack, line 110:
```suggestion
        """Unregister all commands in a manifest"""
```

tlambert03:
> I wonder if you think it'll be useful outside of testing?

I do!  This PR started as a `_testsupport` module with a `PluginFixture` class... and as I realized the generality of it, I pulled it out into something different.  It has a pattern that is not dissimilar to #75 in terms of associating a callable with a command and a contribution.  As it currently stands, I think it's a relatively flexible base to use in many scenarios.  (If you think a name other than `TemporaryPlugin` would help indicate that, I'm open :)

tlambert03:
perhaps `DynamicPlugin`?

tlambert03:
ok to merge if I change name to DynamicPlugin @nclack ?

nclack (approved):

nclack:
> ok to merge

Yes. Meant to approve before anyway.

PR #132, opened 2022-03-16
Add ci to test all plugins on PR label

tlambert03:
This adds a new CI file that will run tests declared in `tests/test_all_plugins.py` on every known plugin.  Need to iterate a few times here to check if it's working

(this includes #124, for the sake of testing ... but that will be either merged before this, or removed before merge here)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,6 @@ on:
 
 jobs:
   test:
-    if: ${{ github.event.label.name != 'test-all-plugins' }}
     name: ${{ matrix.platform }} (${{ matrix.python-version }})
     runs-on: ${{ matrix.platform }}
     strategy:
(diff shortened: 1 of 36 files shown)


tlambert03:
just a bit of data from the tests here.  21 out of the 143 plugins tested actually need the `__npe1shim__` thing at all.  all the rest have direct `python_names` for all their assets

tlambert03:
and... the current object recovery behavior in #124 appears to be working for every one of those 21 plugins

.github/workflows/ci.yml
```diff
@@ -43,6 +43,7 @@ jobs:
           fail_ci_if_error: true
 
   test_napari:
+    if: ${{ github.event.label.name != 'test-all-plugins' }}
```
nclack, line 46: It looks like this triggers for prs with the test-all-plugins label.  Is that right?

nclack (approved):
I keep having trouble looking at the reports from the failing plugins (github times out) but this looks good to me.

nclack:
should this get merged?

tlambert03:
no need, this is mostly just for testing...

PR #134, opened 2022-03-16
Fix ci for testing all plugins

tlambert03:
sorry for the noise, GitHub acting weird with this workflow.  adding a very basic test to see if I can get plugins installing

### DIFF
--- a/tests/test_all_plugins.py
+++ b/tests/test_all_plugins.py
@@ -1,0 +1,22 @@
+import os
+
+import pytest
+
+PLUGIN = os.getenv("TEST_PACKAGE_NAME")
+if not PLUGIN:
+    pytest.skip("skipping plugin specific tests", allow_module_level=True)
+
+try:
+    from importlib import metadata
+except ImportError:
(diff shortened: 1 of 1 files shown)


tests/test_all_plugins.py
```diff
@@ -0,0 +1,33 @@
...
+@pytest.fixture
+def entry_points():
+    d = metadata.distribution(str(PLUGIN))
+    return [
+        ep for ep in d.entry_points if ep.group in ("napari.plugin", "napari.manifest")
+    ]
+
+
+def test_plugin_has_entry_points(entry_points):
+    if PLUGIN not in {"napari-console", "napari-error-reporter"}:
```
nclack, line 25: why not error reporter?

nclack (approved):
lgtm

tests/test_all_plugins.py
tlambert03: Because like console, it doesn't have any hook implementations.  It's a "blessed" plugin with direct injection logic in napari

PR #142, opened 2022-04-01
Push stack=... through all the reader internal API, take II

Carreau:
There is a tiny bit more we can do once
napari/napari#4130 is merged and we assume a
minimal version of napari that only passes list of str to npe2 and give
it the stack=parameter as well, but it will take a couple releases
for this to be the case

Redo of #109 (was too complicated to rebase), so remade in a separate PR


tlambert03 (approved):
looks good to me, thanks @Carreau.  Can you just fixup the coverage issues?  (we're holding to 100% coverage in this repo)

PR #145, opened 2022-04-03
Turn off npe1 discovery by default

tlambert03:
This turns off discovery of npe1 plugins by defaulting, instead accepting an argument to `PluginManager.discover()` to turn on the shim.  this will make it a bit easier for napari to control this behavior, and allow us to release the current main without breaking napari


tlambert03:
@nclack, if you have a moment to review, this will be necessary to let us cut a new npe2 release from `main` without affecting the way napari discovers

tlambert03:
pre-commit test failure is unrelated, and due to an issue with click

nclack (approved):
looks good. thanks!

PR #149, opened 2022-04-05
Add PluginManifest.properties for cross-plugin configuration

### LINKED ISSUE #136: Need way for plugins to communicate about shared or singleton resources - possibly on napari startup
@gselzer raised an important point [on zulip](https://napari.zulipchat.com/#narrow/stream/309872-plugins/topic/Plugin.20Eager.20Initialization/near/276194311).  Here is a specific challenge for plugins using the JVM, but it hints at a broader challenge for plugins that need to share some external initialization: 

> from @gselzer
> Our use case for napari-imagej centers around JVM startup; if multiple plugins want to add JARs to the java runtime before it starts up, they would need a way to do that before any of those plugins actually start the JVM. napari-imagej does its best to defer JVM startup, but if another plugin wants to add additional JARs or if another plugin starts the JVM before napari-imagej does, then one plugin will be unable to run. Ideally, any plugin that wants JARs included in the runtime should be able to declare them in some eager initialization that could run at e.g. plugin discovery time.

@ctrueden proposed something like "kernel space" plugins (such as a `napari-jvm` plugin) that act as aggregators for related functionality, and resources that must be shared:
> One way to protect napari a bit more from rogue plugins could be to have these "kernel-space" plugins act as aggregators for related functionality. What I'm envisioning for Java-dependent plugins would be that you'd have a "Java plugin" for napari, which has some napari-startup-time initialization taking care of preparing the Java runtime. Then the actual Java-dependent plugins like napari-imagej would NOT have startup-time initialization, but rather use a new java_libraries: YAML field or something, which the Java plugin itself consumes. So you'd have a situation then where Java-dependent plugins could declare metadata that affects how the Java plugin prepares the Java runtime, but in a safer way than giving those plugins direct eager code execution.

I like that general idea, as it's extensible for arbitrary other resources.  One question becomes whether `napari-jvm` would be in some way a "blessed" plugin – that is, how does everyone agree on what the kernel-space aggregator is?  And does `npe2` need to add the `java_libraries` field to the npe2 schema?  or does `napari-jvm` register a key during it's special on-startup activation event that other plugins can declare support for?

thanks @gselzer and @ctrueden for raising.  Curious to hear @nclack's thoughts on this particular issue.

gselzer:
This PR introduces the contribution point `properties`, designed to allow plugins to contribute to global state before any individual plugin starts up.

This work is motivated by the napari-imagej plugin, which runs the JVM under the hood. This plugin depends on a set of JARs to run; if another plugin were to start the JVM *before* napari-imagej would,  napari-imagej would not be able to add its JARs to the classpath and would thus be unrunnable.

By declaring all JARs as properties in this new contribution point, JVM-using plugins can install *all* jars needed by all discoverable plugins, allowing all to run.

@tlambert03 [pointed out](https://github.com/imagej/napari-imagej/issues/16#issuecomment-1085789676) that there is already a `configuration` contribution point, however it is my impression that `configuration` and `properties` are designed to do different things:
* `configuration` is designed to solve user-selected *plugin-specific* configuration behavior at plugin startup. In other words, the value solved by `configuration` is not known at napari startup.
* `properties` is designed to solve developer-selected *global* configuration behavior during napari startup. In other words, the value solved by `properties` is known at napari startup.

If I am wrong on this, I am happy to work towards a solution that combines the two. 

I'd love to hear all opinions from @tlambert03 and the napari team more broadly. As long as we can solve the problem of #136 without becoming a burden on the user, I am happy to make any changes.

This PR closes #136.


tlambert03:
just for the sake of completeness, can you show an concrete plugin manifest using `properties` and how npe2 and/or `napari-imagej` (or scyjava, or anything in the loop) would use the properties?  i.e. what sequence of events would occur at launchtime?

gselzer:
> just for the sake of completeness, can you show an concrete plugin manifest using `properties` and how npe2 and/or `napari-imagej` (or scyjava, or anything in the loop) would use the properties? i.e. what sequence of events would occur at launchtime?

Sure, my (maybe naive) plan was to append napari-imagej's `napari.yml` to be:
```yml
name: napari-imagej
display_name: napari-imagej
properties:
  jvm_classpath:
    - 'io.scif:scifio:0.43.1'
contributions:
  commands:
    - id: napari-imagej.func
      python_name: napari_imagej.widget:ImageJWidget
      title: Run ImageJ2 commands
  widgets:
    - command: napari-imagej.func
      display_name: ImageJ2
```

then, in my plugin, I can call:

```python3
import logging
import imagej
from scyjava import config, jimport
from napari.plugins import plugin_manager

# -- LOGGER CONFIG -- #

_logger = logging.getLogger(__name__)
_logger.setLevel(logging.DEBUG)  # TEMP

# -- IMAGEJ CONFIG -- #

# TEMP: Avoid issues caused by https://github.com/imagej/pyimagej/issues/160
config.add_repositories(
    {"scijava.public": "https://maven.scijava.org/content/groups/public"}
)
config.add_option(f"-Dimagej.dir={os.getcwd()}")  # TEMP

for endpoint in plugin_manager.properties("jvm_classpath"):
    config.endpoints.append(endpoint)

_logger.debug("Completed JVM Configuration")
```

Is this what you were looking for @tlambert03?

tlambert03:
yeah, thanks.  need to have a think on it.  busy this week.  will have a look next week and curious to hear @Carreau's thoughts too

gselzer:
Sure thing.

It's a significant addition so I think it warrants serious review/consideration. Happy to work with you guys to revise.

gselzer:
Bumping for visibility

tlambert03:
Hi @gselzer, thanks for your patience, and thanks so much for taking a crack at this PR.

We chatted about this at the plugin dev meeting yesterday.  In short, I'm not sure that a new "properties" field is the best approach here.  The biggest issue here was the open-endedness of `Manifest.properties`.  Note that it would be the first and only "dict" field in the whole manifest, which points to the fact that one could really put anything in there.  And since we really only have one possible use case in mind (the case of `jvm_classpath`), I'm not sure it warrants such an open-ended field that could lead to abuse/confusion.

Because this case of a shared global singleton is unlikely to have _too_ many other similar cases in the near future, we're inclined to just provide you with a `jvm_classpath` field directly (rather than just the "convention" of using `'jvm_classpath'` inside of `properties`.  But I'd still like to have a better picture in my head about how this is going to work with/for scyjava.  The example you gave above ...

```py
for endpoint in plugin_manager.properties("jvm_classpath"):
    config.endpoints.append(endpoint)
```

still seems to have the issue of "whoever gets to the JVM first wins", (though it at least provides a place for everyone to share their needs).  As a side question, why can't we use [`jpype.addClassPath`](https://github.com/jpype-project/jpype/blob/master/jpype/_classpath.py#L27) to add these things after the JVM has been initialized?

If having a `jvm_classpath` field directly in the manifest would solve this, we're definitely open to it, but I think I still need a clearer picture of how this would all play out with scyjava and other packages.  Would you and @ctrueden perhaps have time to zoom about this?  I suspect it would be much easier to come up with something that way.

gselzer:
> And since we really only have one possible use case in mind (the case of `jvm_classpath`), I'm not sure it warrants such an open-ended field that could lead to abuse/confusion.

Fair point. I figured that its open-endedness was a benefit, but yes, you are right that maybe the use cases do not warrant the scope of this solution.

> As a side question, why can't we use [jpype.addClassPath](https://github.com/jpype-project/jpype/blob/master/jpype/_classpath.py#L27) to add these things after the JVM has been initialized?

That...is a fantastic question. Maybe I misunderstood something @ctrueden said about adding classes after JVM startup. The following works on my machine, after downloading a standalone JAR like [parsington](https://github.com/scijava/parsington):
```zsh
(scyjava) : gselzer@gselzer-OptiPlex-5090 ~/code/scijava/scyjava (master) [scyjava] »
python                                                                                                                                                                                                                                                  #  0 {2022-04-21 9:11:25}
Python 3.10.4 | packaged by conda-forge | (main, Mar 24 2022, 17:39:04) [GCC 10.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import scyjava
>>> scyjava.start_jvm()
>>> scyjava.config.add_classpath('/home/gselzer/Downloads/parsington-3.0.0.jar')
>>> scyjava.jimport('org.scijava.parsington.eval.Evaluator')
<java class 'org.scijava.parsington.eval.Evaluator'>
>>> 

```

*Note that `scyjava.config.add_classpath` just calls `jpype.addClassPath` under the hood*

I think the problem is that `scijava.config.endpoints`, the preferred way of adding JARs to the classpath, is *only* added to the classpath in [`scyjava.start_jvm()`](https://github.com/scijava/scyjava/blob/67762c8f7a4a5666a3bb06e0cfac9ba97dcbb3d4/src/scyjava/__init__.py#L110). `scijava.config.endpoints` is preferred as it also adds the dependencies of all entries in the list; you'd have to resolve dependencies yourself if you use `scyjava.config.add_classpath`

I am starting to wonder why we need napari's help on this myself :laughing:

@ctrueden why can't we just add more options (namely endpoint support/dependency resolution) to `scyjava.config.add_classpath`?

tlambert03:
since I don't think we're going to go with the `PluginManifest.properties` approach specifically. I'm going to close this PR.  Thanks so much for the concrete example @gselzer.  let's keep up the conversation in #163

PR #150, opened 2022-04-05
add v0.3.0 changelog

tlambert03:


nclack (approved):
Looks great

tlambert03:
releasing v0.3.0.rc0 now... @Carreau, or @nclack, if you want to check anything on the napari side before i release lemme know

nclack:
Before approving I checked that it wouldn't apply the shim (respecting `include_npe1`).  Just double checked again and it looks fine.  I basically just went through the workflow of installing `napari-animation`, checking that it worked, and then restarting napari and rechecking.

Carreau:
If you do an rc do we/should we install on napari side with `--pre` ?  I don't hink it already does it.

tlambert03:
yeah, I guess that would be the way to test it.

Carreau:
Actually there is already a --pre test, and it should open an issue when failing, except the last auto opened issue is from a year ago and was never closed. 

So I think we can go ahead with a rc0

tlambert03:
whoops sorry, I failed to push the tag yesterday!  just pushed

PR #151, opened 2022-04-06
Auto update changelog workflow

Carreau:
I'm not a ruby person I didn't managed to get the github changelog update gem to work, so why not make it a workflow that can be ran by a press of a button ?

### DIFF
--- a/.github/workflows/update_changelog.yml
+++ b/.github/workflows/update_changelog.yml
@@ -1,0 +1,27 @@
+name: Update Changelog
+on:
+  workflow_dispatch:
+    inputs:
+      next_tag:
+        description: "Next version tag (`vX.Y.Z`)"
+        required: true
+
+jobs:
+  changelog:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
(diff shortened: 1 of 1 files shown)


tlambert03:
this would be great.  it will need the `with: futureRelease` to change each time

tlambert03:
one additional detail here.  when I run this locally, I check the result for any PRs that didn't have labels on them (they get categorized simply as "merged commits").  Do you know if the `create-pull-request` action will let us update a pull request if we re-run this action given the same futureRelease?

Carreau:
> Do you know if the `create-pull-request` action will let us update a pull request if we re-run this action given the same futureRelease?

```
The default behaviour of the action is to create a pull request that will be continually updated with new changes until it is merged or closed. Changes are committed and pushed to a fixed-name branch, the name of which can be configured with the branch input. Any subsequent changes will be committed to the same branch and reflected in the open pull request.
```

Carreau:
I added a workflow input. for "futureRelease"
An alternate possibility is to always run this on master (or run it say once a day) to always have PR with the updated changelog.

PR #159, opened 2022-05-07
Map menu locations from manifest to MenusContribution keys

sofroniewn:
This PR maps menu locations provided in the manifest to MenusContribution keys.

We currently support two menu locations `/napari/layer_context` and `/napari/plugins`. Both of these must be specified exactly like this in the manifest. We will then map them to keys `napari__layer_context` and `napari__plugins` in the `MenusContribution` that napari can then consume.

In the future we could expand the menu locations supported by napari keeping the `/napari` syntax. We could also add support for contributing to menu items from other plugins - though that is not provided in this PR.

This work was done with @kne42 and @tlambert03 at the napari hackathon. cc @nclack @Carreau

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -74,5 +74,5 @@ def test_deep_update():
 
 
 def test_path_to_key():
-    assert 'napari__plugins' == path_to_key('/napari/plugins')
-    assert 'mysubmenu' == path_to_key('mysubmenu')
+    assert "napari__plugins" == path_to_key("/napari/plugins")
+    assert "mysubmenu" == path_to_key("mysubmenu")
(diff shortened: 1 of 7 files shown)


tests/test_utils.py
```diff
@@ -71,3 +71,8 @@ def test_deep_update():
 
     deep_update(a, b, copy=False)
     assert a == {"a": {"b": 1, "d": 4, "c": 3}, "e": 2, "f": 0}
+
+
+def test_path_to_key():
+    assert "napari__plugins" == path_to_key("/napari/plugins")
```
kne42, line 77: you need a test case for more than 2 backslashes eg `/napari/foo/bar`

sofroniewn:
Superseded by #160

PR #160, opened 2022-05-07
Allow arbitrary menu locations in npe2

sofroniewn:
This PR is an alternative to #159 which allows for arbitrary menu locations in npe2. The validation would now happen in napari, or in subsequent PRs. The main change is to drop `MenusContribution`, and instead just have a dict, which allows us to use keys that are strings containing `/` and so we don't need to do any menu coercion.

This PR came from recommendations of @nclack and @tlambert03 after reviewing #159 in person together at the napari hackathon. cc @kne42 @Carreau


kne42 (approved):
much much easier to work with, thank you

npe2/manifest/contributions/_submenu.py
```diff
@@ -1,13 +1,16 @@
...
 from pydantic.fields import Field
 
+from .. import _validators
 from ._icon import Icon
+from ._menus import MenuItem
 
 
 class SubmenuContribution(BaseModel):
     id: str = Field(description="Identifier of the menu to display as a submenu.")
+    _valid_id = validator("id", allow_reuse=True)(_validators.command_id)
```
tlambert03, line 13: 👍

tlambert03 (approved):
lovely

PR #161, opened 2022-05-08
Add list of allowed napari menu locations to npe2

sofroniewn:
This PR adds a list of allowed menu locations to npe2 that plugins can populate with commands or their own submenus. At this point, plugins will not be allowed to contribute to submenus created by other plugins.

The aim of this approach is to ensure that plugins have a reasonable place to put things and users have a reasonable way to find things, but without creating a complex web of interdependencies between plugins.

The initial set of approved locations is taken from places that already exist in napari and the list of locations used in the [napari-tools-menu](https://github.com/haesleinhuepf/napari-tools-menu) example from @haesleinhuepf

@tlambert03 @nclack @sofroniewn @kne42 discussed having this list in napari or in npe2, and decided to put it in npe2 to start as it means that we can validate plugins before they hit napari and evolving this list corresponds to evolving the npe2 interface. The list really defines the valid keys that npe2 recognizes.

@neuromusic might be interested in looking at this initial list and thinking about how we might want to bring it in line with the categories listed in the [napari-hub](https://www.napari-hub.org/). Another nice thing about having this list in npe2 instead of napari is that we can make better use of it on the hub.

Very curious what @jni @DragaDoncila @haesleinhuepf @psobolewskiPhD think


nclack (approved):
lgtm!

npe2/manifest/contributions/_contributions.py
```diff
@@ -27,5 +27,16 @@ class ContributionPoints(BaseModel):
     # configuration: Optional[JsonSchemaObject]
     # keybindings: Optional[List[KeyBindingContribution]]
 
+    @validator("menus")
+    def _check_napari_menus(cls, v):
+        valid_names = [n.key for n in napari_menus]
```
tlambert03, line 32: minor nit, this should be a set and probably put at module level (since it won't change  within a session:
```python
VALID_NAMES = {n.key for n in napari_menus}
```

tlambert03 (approved):
tiny nit comment, otherwise yay

npe2/manifest/contributions/_contributions.py
```diff
@@ -1,9 +1,9 @@
 from typing import Dict, List, Optional
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, validator
 
 from ._commands import CommandContribution
-from ._menus import MenuItem
+from ._menus import MenuItem, napari_menus
```
kne42, line 6: what do you think about renaming `MenuItem` to `MenuContribution` to match with the other fields?

npe2/manifest/contributions/_menus.py
```diff
@@ -1,27 +1,26 @@
...
+    Menu("/napari/layer_context", "Process Layer"),
+    Menu("/napari/layer_context/projections", "Make Projection"),
+    Menu("/napari/layer_context/convert_type", "Convert datatype"),
+    Menu("/napari/tools/acquisition", "Acquisition"),
+    Menu("/napari/tools/classification", "Classification"),
+    Menu("/napari/tools/measurement", "Measurement"),
+    Menu("/napari/tools/segmentation", "Segmentation"),
+    Menu("/napari/tools/transform", "Transform"),
+    Menu("/napari/tools/utilities", "Utilities"),
+    Menu("/napari/tools/visualization", "Visualization"),
```
haesleinhuepf, line 22:
```suggestion
    Menu("/napari/layer_context", "Process Layer"),
    Menu("/napari/layer_context/projections", "Make Projection"),
    Menu("/napari/layer_context/convert_type", "Convert datatype"),
    Menu("/napari/tools/acquisition", "Acquisition"),
    Menu("/napari/tools/classification", "Classification"),
    Menu("/napari/tools/filters", "Filters"),
    Menu("/napari/tools/measurement", "Measurement"),
    Menu("/napari/tools/segmentation", "Segmentation"),
    Menu("/napari/tools/projection", "Projection"),
    Menu("/napari/tools/transform", "Transform"),
    Menu("/napari/tools/utilities", "Utilities"),
    Menu("/napari/tools/visualization", "Visualization"),
```
Two things (open for discussion):
1. Denoising / background-removal / etc. are calling for a "Filters" menu. It could also be called "Process" like in other related software.
2. I'm adding the "Projection" menu to the main menu in parallel to the right-click menu. Reason: Users don't see that that right-click menu exists. The main menu is visible. Thus, projections become more accessible. Alternatively: Projections and transforms go well together in one menu.

haesleinhuepf (commented):
Hi @sofroniewn , hi all,

exciting PR! So happy to see menus evolving. As you know, I was experimenting with a "Tools" menu and I'd like to share typical feedback I got in this explorative phase.
* Alphabetical order is not intuitive. Consider ordering in a typical workflow order:
  * Acquisition
  * Transform / project
  * Filter
  * Segmentation
  * Measurement
  * Classification
  * Visualization
  * Utilities  
* In our prototype we had multiple menus for the different kind of "Filters". New users are confused by that. Often, beginners don't know the technical difference between "Noise removal" and "Background removal". So putting those together in one menu makes sense.
* Similarly, we have a separation between "Binarization" and "Labeling". Users don't care about this difference. Having one unified "Segmentation" menu makes a lot of sense.
* Going from an image to a labels layer via "Segmentation" is clear and obvious. In the prototype we additionally have two menus "Points" and "Surfaces" for going back and forth betwen `labels` layers, `points` layers and `surface` layers. It is super hard for the users to find anything in these menus. If anyone has a good idea of in which menu to put point- and surface- processing / analysis steps, I'd be very curious to learn more!
* Our current tools menu is quite long. I _think_ you are aiming for a shorter menu and fully support that. Ideas: Put "Projection / Transformation" together. Also consider combining "Measurement / Visualization".

Please keep me posted. As soon as there is a prototype implementation available, I would love to try it out!

Fantastic work! :-)

Best,
Robert

haesleinhuepf:
One more thought: Shall we create a NAP and discuss the menu structure in general?

npe2/manifest/contributions/_contributions.py
tlambert03: This is kind of a special case, the "real" `MenuContribution` in this is probably the dict: `Dict[str, List[MenuItem]]`.  But to keep it consistent we could make a type alias over in _menus and use it here?

```
MenuContribution = Dict[str, List[MenuItem]]

class ContributionPoints(BaseModel):
    ...
    menus: Optional[MenuContribution]
```

?

npe2/manifest/contributions/_contributions.py
kne42: +1 this will make it easier to access from the napari side as well

npe2/manifest/contributions/_menus.py
```diff
@@ -1,27 +1,26 @@
...
-#     ),
-#     Menu(
-#         key="layers__context", id=1, description="The layer list context menu"
-#     ),
-# ]
+Menu = namedtuple("Menu", "key description")
+
+# a list of valid napari menu locations that plugins can contribute too.
+# keys provided in the plugin manifest must match these strings
+napari_menus = [
```
kne42, line 12: what about using a dict instead of a nx2 array? this would make it easier to get just the paths since u could go `napari_menus.keys()`

npe2/manifest/contributions/_menus.py
```diff
@@ -1,27 +1,26 @@
...
-# napari_menus = [
-#     Menu(key="command_pallete", id=0, description="The Command Palette"),
-#     Menu(
-#         key="layers__context", id=1, description="The layer list context menu"
-#     ),
-#     Menu(
-#         key="layers__context", id=1, description="The layer list context menu"
-#     ),
-# ]
+Menu = namedtuple("Menu", "key description")
```
sofroniewn, line 8: @kne42 should we make the "id, description" - trying to match what you did here https://github.com/napari/napari/pull/4497/files#diff-9ea32b5da7975500739c6eb2f806a3ca84c916c445cf4e5051b6e9d5fe903312R8

sofroniewn (commented):
Thanks for that feedback @haesleinhuepf - I think having just transform is fine for now, we can add projection later if we want. I've also added some "order" numbers for making menu in right order.

So far there has been quite a bit of consensus around this approach and list of options, so @jni and I don't feel like we need a NAP right now, but if anyone has strong feelings like this is not the direction we should go we could write one.

npe2/manifest/contributions/_menus.py
```diff
@@ -1,26 +1,28 @@
...
-from typing import Optional, Union
+from typing import Dict, List, Optional, Union
 
 from pydantic import BaseModel, Field
 
 from ..utils import Executable
 
-Menu = namedtuple("Menu", "key description")
+
+Menu = namedtuple("Menu", "key description order")
```
kne42, line 8: `'path title order'` makes more sense to me naming wise

nclack:
@sofroniewn Could you take one more look at the review feedback?  Also, I believe this is missing a test that covers what happens when a contribution provides a menu to an invalid top-level location.

npe2/manifest/contributions/_contributions.py
```diff
@@ -21,11 +23,21 @@ class ContributionPoints(BaseModel):
...
+    menus: Optional[MenuContribution] = Field(None, hide_docs=True)
     submenus: Optional[List[SubmenuContribution]] = Field(None, hide_docs=True)
 
     # configuration: Optional[JsonSchemaObject]
     # keybindings: Optional[List[KeyBindingContribution]]
 
+    @validator("menus")
+    def _check_napari_menus(cls, v):
+        for key in v.keys():
+            if key not in VALID_MENUS_NAMES:
```
tlambert03, line 35: I'm also not sure if this is still going to work if we need to go back on our plan to disallow plugins to populate other plugin's menus.  Quick reminder: this came up on zulip recently, where someone wanted to be able to have, for example, their own "group" menu (e.g. `Smith Lab Plugins`) ... and populate it across their multiple plugins.

If we state that only our known menu keys are valid, that won't work

nclack: Ah, I forgot about that. What if we just restricted the top-level menus, but allowed things to be more flexible for sub-menus?

tlambert03: but how do you determine what is top level and what is not?   put another way: how do you envision plugin B contributing to plugin A's menu if not by adding their menu key to the list of menus

tlambert03: i know this is complicated/tricky stuff :joy: ... if you have a vision for how to restrict/validate this _at manifest ingestion time_ without also breaking the ability to add to arbitrary plugin's menus.  Can you explain it in a bit more detail (with examples)

sofroniewn: > I'm also not sure if this is still going to work if we need to go back on our plan to disallow plugins to populate other plugin's menus.  Quick reminder: this came up on zulip recently, where someone wanted to be able to have, for example, their own "group" menu (e.g. Smith Lab Plugins)

ah right - i might loop @chili-chiu into this discussion too to get another opinion here too. i'm not sure what's best, but that's an understandable request

nclack: Using the sample:
```yaml
  menus:
    /napari/layer_context:
      - submenu: my-plugin.mysubmenu
      - command: my-plugin.hello_world
  submenus:
    - id: my-plugin.mysubmenu
      label: My SubMenu
      contents:
        - command: my-plugin.another_command
```

Another plugin could do:
```yaml
  menus:
    /napari/layer_context:
      - submenu: another-plugin.mysubmenu
  submenus:
    - id: another-plugin.mysubmenu
      label: My SubMenu
      contents:
        - command: another-plugin.super_sweet_command
```

I'm imagining the result in napari would be:
```
<layer context menu>
  - My SubMenu 
     - <command title for my-plugin.another_command>
     - <command title for another-plugin.super_sweet_command>
```

psobolewskiPhD:
Late on this, but I *really* like the idea of a task/categorical plugin or tools menu, rather than it being `napari-foo` and `napari-bar` all the way down.
At the same time I also really like the idea of curated collections/meta-plugins/whatever-you-want-to-call-them. Again, anything to make it more organized.
So thumbs up for this, can't wait to see how it works beyond the @haesleinhuepf-ecosystem.

npe2/manifest/contributions/_contributions.py
sofroniewn: @tlambert03 could we
1) go with the original menus/submenus construction approach
2) allow plugin A to contribute to submenu defined by plugin B
3) have approved locations like `/napari/tools/classification` that populate top of menu
4) put custom submenus like `napari/tools/Smith Lab` below a divider (like VScode had those submenu section, remember?)

this would be very flexible, but at least encourage use of our organizational system. thoughts?

tlambert03: for steps 3/4 ... exactly where an item appears within a menu is yet to be determined.  So, the answer is kind of "yes, nothing so far limits us from doing that".  But I bet we'll come into issues with it later.  This comes back to the idea of menu groups and ordering that we discussed at the hackathon.  Here's how vscode deals with it: https://code.visualstudio.com/api/references/contribution-points#Sorting-of-groups

Anyway, I see steps 3/4 (the "where" question) as slightly orthogonal to steps 1/2 (the "what" question).  We can revisit the "where" question later (I think?)

nclack: > I'd propose we go back to not having submenus.contents at all

Good point.

I'll try to restructure for the vscode-like design (which I do like).  I think the basic idea is the same though: Join submenus in napari based on the label - I'm imagining that's a napari-side behavior.

So the sample plugin becomes:
```yaml
  menus:
    /napari/layer_context:
      - submenu: my-plugin.mysubmenu
      - command: my-plugin.hello_world
    my-plugin.mysubmenu:
        - command: my-plugin.another_command
  submenus:
    - id: my-plugin.mysubmenu
      label: My SubMenu
```

And the other plugin is:

```yaml
  menus:
    /napari/layer_context:
      - submenu: another-plugin.mysubmenu
    another-plugin.mysubmenu:
      - command: another-plugin.super_sweet_command
  submenus:
    - id: another-plugin.mysubmenu
      label: My SubMenu
```

Same output as before
```
<layer context menu>
  - My SubMenu 
     - <command title for my-plugin.another_command>
     - <command title for another-plugin.super_sweet_command>
```

tlambert03: correct... but I think you left out the key example there, which would be...

The second plugin adds an item to the first plugin's menu.  (whether the second plugin adds a new submenu is fine, but not the critical thing we're trying to determine here)

```yaml
  menus:
    /napari/layer_context:
      - submenu: another-plugin.mysubmenu
    my-plugin.mysubmenu:
      - command: another-plugin.super_sweet_command
```

nclack: I'm thinking a plugin only refers to its own menu ids; it's self-contained. It adds to another plugin's menu only by the coincidence of using the same label ("My SubMenu" above).

In your example, if "my-plugin.submenu" moved then "another-plugin.super_sweet_command" would move along with it. In my proposal, it wouldn't.  Does that make sense?  What's the desired behavior?

tlambert03: > I'm thinking a plugin only refers to its own menu ids; it's self-contained. It adds to another plugin's menu only by the coincidence of using the same label ("My SubMenu" above).

so, you're essentially just saying that "label" _is_ the identifier?   that feels icky.  The label is how the menu _appears_, it shouldn't be the globally unique identifier.  What if two labs just happen to use the same name for their submenu, but one wants to place theirs somewhere and the other wants to place it somewhere else?  In your example, they just "happened" to both be in `/napari/layer_context:` ... but what if they weren't?  How would the second party indicate which of the two menus they wanted?  (answer: the globally unique identifier that we asked the submenu creator to create in the first place :joy:).  That one's a hard no for me :)  labels make bad identifiers.

> In your example, if "my-plugin.submenu" moved then "another-plugin.super_sweet_command" would move along with it. 

That's by design.  Remember, despite the name, a submenu doesn't "live" anywhere specific.  In fact, it can live in multiple places (and in vscode they very often do).  Don't think of it as "this submenu always lives in layers context".  Think of it as "my lab offers a menu... if you'd like to participate in it, here's our unique id".

nclack: > > In your example, if "my-plugin.submenu" moved then "another-plugin.super_sweet_command" would move along with it.
> 
> That's by design.

Ah, I was wondering about that. If that's a requirement then my thing definitely is a no-go. 

> labels make bad identifiers

Agreed. The identifier determines how submenus get attached to different places, but when they do it can happen that there are two things with the same label. We could choose to combine them at that point. 

In any case, I've probably made this thread too long :) Happy to discuss more over zulip etc.

jni: > so, you're essentially just saying that "label" is the identifier? that feels icky.

Even though the thread might be too long :joy:, I just want to add my strong :+1: here, I instantly wanted to go in and change the id in `another-plugin` to `my-plugin.mysubmenu` so that it would match where it was going. If two labels are the same, but the ids are different, I would expect two different labels, perhaps with some note to disambiguate them.

sofroniewn: > I instantly wanted to go in and change the id in another-plugin to my-plugin.mysubmenu so that it would match where it was going. If two labels are the same, but the ids are different, I would expect two different labels, perhaps with some note to disambiguate them.

could we ever get in a situation though then where a plugin references a menu location that doesn't exist? say if the other plugin where that location was originally defined wasn't installed? if we can always figure out something sensible to do in that case then i'd feel much better about this (which is what i liked about nathan's approach)

tlambert03: > could we ever get in a situation though then where a plugin references a menu location that doesn't exist? say if the other plugin where that location was originally defined wasn't installed?

absolutely. so we'll need to handle a missing menu id gracefully.  I'll look into how vscode does it

tlambert03:
closing since #160 is reverted.  will follow up with another PR, and perhaps some discussion over in the napari repo

PR #164, opened 2022-05-22
Add `plugin_manager` module for global singleton convenience 

tlambert03:
Inspired by some conversations with @kne42 at the hackathon, this PR adds a new module that serves as a convenience for accessing public methods on the global `PluginManager.instance()` singleton.

Everything is lazy, so importing the module does nothing to instantiate the singleton.  I avoided fully autogenerating the module so as to retain type hinting and basic docs (though the full docstrings with parameters will remain in the actual PluginManager methods).

This would make a lot of the lines in `napari/plugins/_npe2.py` and elsewhere simpler.  for example:

```python
pm = npe2.PluginManager.instance()
for plugin_name, contribs in pm.iter_sample_data():
   ...

# becomes
for plugin_name, contribs in npe2.pm.iter_sample_data():
    ...
```

The only downside is that the `pm.py` module needs to be kept manually updated (but there's a test to let us know when we need to).

thoughts @kne42, @nclack?


nclack:
I'm not a big fan of using `pm` as a name - I tend to want to spell things out rather than abbreviate.  What do you think about making the singleton transparent so things look like: `npe2.iter_sample_data()`?

tlambert03:
I'm not married to `pm`.  Doing it in the root namespace is definitely aesthetically appealing, but it scares me to use that namespace 😂!  I _guess_ I could live with `from npe2 import plugin_manager as pm` or something like that?

nclack:
root namespace scares me too 👻 `plugin_manager` looks great.

tlambert03:
Will merge this tommorrow with new module name.  Unless @nclack you think we shouldn’t have it at all? (You never commented on that)

npe2/_plugin_manager.py
```diff
@@ -471,9 +484,17 @@ def iter_compatible_readers(
     def iter_compatible_writers(
         self, layer_types: Sequence[str]
     ) -> Iterator[WriterContribution]:
+        """Iterate over compatible WriterContributions given a sequence of layer_types.
```
nclack, line 487: love all the improved docstrings!

nclack (approved):
I _do_ like the idea of the global `plugin_manager` instance.

PR #165, opened 2022-05-23
Fix potential error with `npe2 cache --list` with uninstalled plugin

tlambert03:
calling `distribution()` on an uninstalled plugin is an error.  This PR just uses `mf.package_version` instead, which will catch the error


nclack (approved):
thanks!

PR #172, opened 2022-06-12
Fix writer order preference

### LINKED ISSUE #129: Any way to allow a plugin control over it's writer priority?
In converting over to using only npe2 interface in napari, I came across an issue where it was very difficult to get `get_writer` to pick the lossless writer by default when writing an image with `write_layer_data_with_plugins`:

```python
        # build in a temporary directory and then move afterwards,
        # it makes cleanup easier if an exception is raised inside.
        with TemporaryDirectory(dir=path) as tmp:
            # Loop through data for each layer
            for layer_data_tuple in layer_data:
                _, meta, _ = layer_data_tuple
                # Get hook caller according to layer type

                # Create full path using name of layer
                full_path = abspath_or_url(os.path.join(tmp, meta['name']))
                # Write out data using first plugin found or named plugin if provided
                written.extend(
                    io_utils.write(
                        path=full_path,
                        layer_data=[layer_data_tuple],
                        plugin_name=plugin_name,
                    )
                )
            for fname in os.listdir(tmp):
                shutil.move(os.path.join(tmp, fname), path)
```

The tiff writer does appear first in the list of single image writers, but there's a few things in iter_compatible writers that mangle the order.  Ultimately, I needed to remove these two lines from `_writer_key` in order to get it to choose the earlier listed `tiff` writer instead of the lossy writer:

```
            # 3. then sort by the number of listed extensions
            #    (empty set of extensions goes last)
            ext_len = len(writer.filename_extensions)

            # 4. finally group related extensions together
            exts = writer.filename_extensions
```

But, obviously, that's just a hack.  @nclack, we didn't particularly anticipate this internal usage, but do you have any thoughts on how we might update our `write_layer_data_with_plugins` to maintain the current behavior while using npe2?

tlambert03:
fixes #129, ensuring that – all else being equal – writers declared earlier in a plugin's manifest have priority over later ones

### DIFF
--- a/npe2/_plugin_manager.py
+++ b/npe2/_plugin_manager.py
@@ -169,7 +169,7 @@ class _ContributionsIndex:
             nbounds = sum(not c.is_zero() for c in writer.layer_type_constraints())
             return (no_ext, nbounds)
 
-        yield from candidates
+        yield from sorted(candidates, key=_writer_key)
 
 
 class PluginManagerEvents(SignalGroup):
--- a/npe2/_plugin_manager.py
+++ b/npe2/_plugin_manager.py
@@ -160,25 +160,15 @@ class _ContributionsIndex:
             else:
                 break
 
-        def _writer_key(writer: WriterContribution) -> Tuple[bool, int, int, List[str]]:
+        def _writer_key(writer: WriterContribution) -> Tuple[bool, int]:
             # 1. writers with no file extensions (like directory writers) go last
             no_ext = len(writer.filename_extensions) == 0
 
             # 2. more "specific" writers first
             nbounds = sum(not c.is_zero() for c in writer.layer_type_constraints())
+            return (no_ext, nbounds)
 
-            # 3. then sort by the number of listed extensions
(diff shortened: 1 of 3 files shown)


npe2/_plugin_manager.py
```diff
@@ -153,32 +153,23 @@ def _get_candidates(lt: LayerType) -> Set[WriterContribution]:
...
+        candidates = list({w: None for _, _, _, w in self._writers})
         for lt in LayerType:
             if candidates:
-                candidates &= _get_candidates(lt)
+                candidates = [i for i in candidates if i in _get_candidates(lt)]
             else:
                 break
 
-        def _writer_key(writer: WriterContribution) -> Tuple[bool, int, int, List[str]]:
+        def _writer_key(writer: WriterContribution) -> Tuple[bool, int]:
```
nclack, line 164: doesn't look like `_writer_key` is being used anymore so this can be removed.

nclack (commented):
thanks! Looks like there's some dead code that should be removed.

PR #173, opened 2022-06-12
Add minor conveniences for DynamicPlugin

tlambert03:
Some stuff I find myself using with DynamicPlugin over in napari:  2 quick access properties, and a "spawn" method: a way to quickly create a second temporary plugin using the same plugin manager as a given dynamic plugin


nclack (approved):
very cool.  lgtm.

PR #174, opened 2022-06-12
Revert menu restriction (#160)

tlambert03:
This reverts #160, which i think we've decided may not be the direction we want to go in after all (see discussion in #161 and related links)... primarily since we likely need plugins to be able to contribute to other plugins menus.

This is also the one thing that's causing our napari test fails.

After this: consider #175


nclack (approved):
thanks. nice to get tests passing.

PR #178, opened 2022-06-13
hide docs again in napari menus (Fix napari docs build)

tlambert03:
forgot to add `hide_docs` in Contributions.menus.

Actually, just realizing that since we use the schema.json from the latest release (not from main), this actually requires a new release to fix building docs on napari.  So this PR adds a new test to make sure that docs are building correctly.  (I can't remember the backstory, but the docs build test only runs on main, and it still for some reason, failed... so this adds another test that should catch it)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ on:
   workflow_dispatch:
 
 jobs:
+
   test:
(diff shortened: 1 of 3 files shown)


tlambert03:
everything but `docs render` is going to fail now until the next release (coming in a minute), so merging this one

PR #182, opened 2022-06-19
Add npe1_shim field to schema

tlambert03:
This adds a new top level `npe1_shim` field to the schema, which will make it possible to determine whether a serialized manifest represents a shimmed npe1 plugin or not.

This also makes sure that all shim manifests have their contributions loaded before serialization 

cc @DragaDoncila

### DIFF
--- a/npe2/manifest/_npe1_adapter.py
+++ b/npe2/manifest/_npe1_adapter.py
@@ -81,13 +81,14 @@ class NPE1Adapter(PluginManifest):
         self._dist = dist
 
     def __getattribute__(self, __name: str):
-        if __name == "contributions" and not self._is_loaded:
+        if __name == "contributions":
             self._load_contributions()
         return super().__getattribute__(__name)
 
     def _load_contributions(self, save=True) -> None:
         """import and inspect package contributions."""
-
+        if self._is_loaded:
+            return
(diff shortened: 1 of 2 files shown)


DragaDoncila (approved):
This looks great to me, thank you @tlambert03 🙏 . Two questions:

- Could we use this in the plugin dialog on the napari side to avoid `npev = 'shim' if 'npe1' in type(manifest).__name__.lower() else 2`? I know with napari #4692 it'll matter less but it may still be useful for users to know whether a plugin is native `npe2` or not.
- Now that we load contributions before serializing, does that mean we can get rid of the call to `index_npe1_adapters` on the napari hub side which we were using to force contribution discovery?

tlambert03:
> Could we use this in the plugin dialog on the napari side to avoid npev = 'shim' if 'npe1' in

yes!

> does that mean we can get rid of the call to index_npe1_adapters on the napari hub side which we were using to force contribution discovery, even though we're exporting to json

yes!

tlambert03:
> does that mean we can get rid of the call to index_npe1_adapters on the napari hub side which we were using to force contribution discovery, even though we're exporting to json

actually, even better @DragaDoncila ... you should just be able to do this now from the cli

```sh
npe2 parse napari-compressed-labels-io > napari.yaml
```

PR #184, opened 2022-06-19
Fail silently when caching throws OSError

DragaDoncila:
Saving to cache with the current `cache_path` could (rarely) throw an `OSError` if the user didn't have permission to write to `/home`. We have run into this on the napari hub when working in a Docker image that only allows us to write to `/tmp`.

Since caching is just a convenience, this PR catches `OSError` on caching and fails silently.


tlambert03 (approved):
🎉 thanks!

PR #185, opened 2022-06-20
Add `npe2 fetch` command to cli to fetch remote manifests

tlambert03:
~cc @DragaDoncila ... just a proof of principle at the moment... but working for a couple challenging ones, like `npe2 fetch napari-omero` (works because it doesn't try to install) and `npe2 fetch napari-clusters-plotter` (works because it creates a temporary install with `napari[all]` ... which is likely why many of the nometa group are failing)~

~The goal here is to essentially try the fastest, easiest, most direct methods first, then fall back to fully installing~

This adds a new `fetch` command to npe2 that will grab a manifest for any napari plugin listed on pypi.  This *includes* npe1 packges, which will be installed into a temporary isolated environment, imported, and indexed into an npe1 shim
```sh
╰─❯ npe2 fetch napari-omero
name: napari-omero
schema_version: 0.1.0
contributions:
  commands:
  - id: napari-omero.widget
    title: Open OMERO widget
    python_name: napari_omero.widgets.main:OMEROWidget
  - id: napari-omero.get_reader
    title: OMERO reader
    python_name: napari_omero.plugins._napari:napari_get_reader
  readers:
  - command: napari-omero.get_reader
    filename_patterns:
    - omero://*
  widgets:
  - command: napari-omero.widget
    display_name: OMERO Browser
```

### DIFF
--- a/.github/workflows/test_all_plugins.yml
+++ b/.github/workflows/test_all_plugins.yml
@@ -42,29 +42,6 @@ jobs:
           miniforge-version: latest
           use-mamba: true
 
-      - id: check_conda
-        run: echo ::set-output name=status::$(curl -s -o /dev/null -w '%{http_code}' 'https://api.anaconda.org/package/conda-forge/${{ matrix.plugin }}')
-
-      # for now, until clesperanto is on conda and can declare it itself
-      - name: openCL deps
-        if: matrix.plugin == 'napari-pyclesperanto-assistant' && steps.check_conda.outputs.status != '200'
-        run: mamba install pyopencl oclgrind
-
-      # these plugins offer napari functionality
-      # but also work in the absence of napari... so probably don't want
-      # to add napari to their deps.
-      - name: install napari
-        if: matrix.plugin == "RedLionfish" || matrix.plugin == "smo"
-        run: mamba install napari
-
-      - name: Install ${{ matrix.plugin }} from conda
-        if: steps.check_conda.outputs.status == '200'
-        run: mamba install ${{ matrix.plugin }} pyqt
-
-      - name: Install ${{ matrix.plugin }} from pip
-        if: steps.check_conda.outputs.status != '200'
-        run: pip install ${{ matrix.plugin }} pyqt5
-
       - name: Install npe2
         run: pip install -e .[testing]
 
--- a/npe2/_fetch.py
+++ b/npe2/_fetch.py
@@ -229,6 +229,7 @@ def _fetch_npe1_manifest(package: str, version: Optional[str] = None) -> PluginM
                 warnings.filterwarnings(
                     "error", message="Error importing contributions"
                 )
+                warnings.filterwarnings("ignore", message="Found a multi-layer writer")
                 mf._load_contributions(save=False)
         return mf
 
--- a/npe2/_fetch.py
+++ b/npe2/_fetch.py
@@ -205,54 +205,52 @@ def isolated_plugin_env(
         sys.path.insert(0, site_pkgs[0])
         try:
             if validate_npe1_imports:
+                # try to import the plugin's entry points
                 dist = metadata.distribution(package)
-
-                npe1_eps: List[metadata.EntryPoint] = []
-                npe2_ep: Optional[metadata.EntryPoint] = None
-                for ep in dist.entry_points:
-                    if ep.group == NPE2_ENTRY_POINT:
-                        npe2_ep = ep  # pragma: no cover
-                    elif ep.group == NPE1_ENTRY_POINT:
-                        npe1_eps.append(ep)
-
-                if npe2_ep is None:
-                    for ep in npe1_eps:
-                        try:
-                            ep.load()
-                        except ImportError:
-                            # if loading contributions fails, it can very often be fixed
-                            # by installing `napari[all]` into the environment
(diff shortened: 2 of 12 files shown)


npe2/manifest/utils.py
```diff
@@ -304,3 +324,211 @@ def merge_contributions(contribs: Sequence[Optional[ContributionPoints]]) -> dic
...
+            raise ValueError(f"{name} does not have version {version}")
+        _releases: List[dict] = data["releases"][version]
+    else:
+        _releases = data["urls"]
+
+    releases = {d.get("packagetype"): d for d in _releases}
+    if packagetype:
+        if packagetype not in releases:
+            version = version or "latest"
+            raise KeyError('No {packagetype} releases found for version "{version}"')
```
DragaDoncila, line 377:
```suggestion
            raise KeyError(f'No {packagetype} releases found for version "{version}"')
```

This should be an f-string no?

npe2/manifest/utils.py
```diff
@@ -304,3 +324,211 @@ def merge_contributions(contribs: Sequence[Optional[ContributionPoints]]) -> dic
...
+    """Context in which a pip specifier is installed and added to sys.path.
+
+    Parameters
+    ----------
+    package : str
+        pypi package name
+    version : str, optional
+        optional package version, by default, latest version.
+    stdout : subprocess._FILE
+        stdout for pip install, by default DEVNULL
```
DragaDoncila, line 398: missing `extra_packages` here

.github/workflows/test_all_plugins.yml
```diff
@@ -42,29 +42,6 @@ jobs:
           miniforge-version: latest
           use-mamba: true
 
-      - id: check_conda
```
DragaDoncila, line 45: Can you clarify why we're removing these checks?

tlambert03:
for our notes... here's the current state of all-plugin tests:


## Maybe a napari problem?

- napari-tomoslice - napari/settings/_appearance.py - module 'npe2.plugin_manager' has no attribute '_themes'
- napari-subboxer - 'npe2.plugin_manager' has no attribute '_themes'

## Package needs to fix manifest

- napari-flowering-apple-tree-pixel-classification - not a valid display_name. (update docs on npe2)
- napari-labels-overlap - errors=[{'loc': ('engine',), 'msg': 'extra fields not permitted' (ignore?)

## Package needs to update dependencies or build

- beetlesafari - "No module named 'napari_pyclesperanto_assistant'"
- napari-IP-workflow - has `skimage` in their dependencies
- napari-mri - "No module named 'napari_mri'" (weird packing error?)
- skeleton-finder: broken package, no python files
- napari-yapic-prediction - ModuleNotFoundError: No module named 'tensorflow'
- nfinder - ModuleNotFoundError: No module named 'matplotlib'
- napari-yolov5 - No module named 'magicgui.events' in __init__

## Package needs to ship manifest in sdist/wheel

- btrack
- napari-stl-exporter

## Might need conda

- napari-pyclesperanto-assistant - clGetPlatformIDs failed: PLATFORM_NOT_FOUND_KHR
- grabber-ift - failed to install
- napari-lazy-openslide - OSError: libopenslide.so.0: cannot open shared object file
- napari-cupy-image-processing - no install
- napari-sdeconv - maybe failed to build?

# Other

- napari-zelda - No module named 'PyQt5'"

npe2/_fetch.py
```diff
@@ -0,0 +1,269 @@
...
+                    mf = PluginManifest.from_file(str(mf_file))
+                    # manually add the package metadata from our distribution object.
+                    mf.package_metadata = PackageMetadata.from_dist_metadata(
+                        dist.metadata
+                    )
+                    return mf
+                elif ep.group == NPE1_ENTRY_POINT:
+                    has_npe1 = True  # pragma: no cover
+    except KeyError as e:
+        if "No bdist_wheel releases found" not in str(e):
```
DragaDoncila, line 85: Am I understanding correctly that if a wheel isn't present we fall back to full install rather than parsing the `sdist`?

DragaDoncila (approved):
Approved, left two questions just for my own understanding.

To support the hub preview page it might be nice if we can pass `fetch` a GitHub repo & branch and have it try fetch the manifest. Currently for the preview page we're relying on install for `npe2` stuff, but before that we were building a wheel and that works for pretty much all plugins. This is definitely not required for this PR though (or really, at all), but it would be great to unify the different mechanisms we have for generating metadata.

npe2/_fetch.py
tlambert03: correct.  I worked on parsing the sdist a bit... but it's a bit of a can of worms.  In a future PR, I do think we could speed up a few npe2 repos that do have manifests, but don't have wheels

.github/workflows/test_all_plugins.yml
tlambert03: Cause I'm doing all of the env building now in the test itself (using `isolated_plugin_env`) ... so it's not needed in the test setup anymore

nclack (commented):
I'm super excited for this.

What do you think about adding logging for the various phases of the fetch. Some of the plugins take a while (e.g. the ones that have to install stuff?) and the logging could help me tell where things were taking time.

PR #186, opened 2022-06-20
Compile plugins using `npe2.implements`

tlambert03:
This would follow on #75, and provide an `npe2 compile` step that would render the full manifest (perhaps during the wheel/sdist build step) by inspecting the package AST.

more docs to follow


npe2/implements.py
```diff
@@ -256,6 +261,15 @@ def visit(
...
         Module name, by default ""
+    visit_comments: bool
+        Whether to try to parse comments for commented out contributions. This
+        lets the plugin avoid depending on npe2 entirely at runtime. by default, True
+
+            # @npe2.implements.writer(
+            #     id="my_single_writer",
+            #     ...
+            # )
+            def my_writer(...): ...
```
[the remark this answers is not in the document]
tlambert03: oh sorry, this should be removed... this PR contained the (now-closed) #189... and I missed this specific parameter when cleaning up.  
I could make `__name__` configurable, but this also works, and keeps all of the IDE type checking intact:

```python
if TYPE_CHECKING:
     from npe2 import implements
 else:
     # create no-op `implements.anything(**kwargs)` decorator
     D = type("D", (), {"__getattr__": lambda *_: (lambda **_: (lambda f: f))})
     implements = D()
```

Carreau:
That sounds reasonable to me. At. alater task we might want to try to list contributions that have not been automatically found to have a decorator, but are in the manifest.

tlambert03:
Marking this ready for review and removing from draft.  I'd like to get this one off my back :joy:.  It works ok, and more tests will come when it gets hooked up to a setuptools plugin in #194 

for now, this should allow `npe2 compile /path/to/repo` to pump out a manifest for that repository using inspection

npe2/_inspection/_setuputils.py
```diff
@@ -82,6 +82,18 @@ def get_package_dir_info(path: Union[Path, str]) -> PackageInfo:
                 name, _, value = val.partition("=")
                 info.entry_points.append(EntryPoint(name.strip(), value.strip(), group))
 
+    # # check for pyproject.toml
```
nclack, line 85: (This comment isn't quite in the right place).
When I test on `napari-animation`, the package name get's pulled out of `setup.cfg` correctly, but overwritten by `None` in the `setup.py` path.

I don't think the logic really needs to change - just a guard needed on the `info.package_name` update.

tlambert03: yeah, thanks.  packages with multiple config files definitely need work.  the guard would definitely be better than the current behavior! :joy:  I also want to do a little more research to learn what setuptools would actually do in the case of, say, all three files.  I believe their priority is 1) pyproject 2) setup.cfg 3) setup.py,  but need to check.  will add guard here

tlambert03: added

npe2/cli.py
```diff
@@ -493,5 +499,28 @@ def cache(
...
+        "--output",
+        exists=False,
+        help="If provided, will write manifest to filepath (must end with .yaml, "
+        ".json, or .toml). Otherwise, will print to stdout.",
+    ),
+    format: Format = typer.Option(
+        "yaml", "-f", "--format", help="Markdown format to use."
+    ),
+):
+    """Compile napari_plugin_engine plugins."""
```
nclack, line 517: I suspect the description of the command isn't quite right.

I'm a little confused about this.  Is this compiling the `@npe2.implements` stuff or is it extracting npe1 hooks?

`_inspection.compile()` ends up calling `find_npe2_module_contributions()` which seems non-npe1.

tlambert03: Correct, this whole implements/compile thing is about autogenerating manifests for npe2 plugins.  Npe1 not supported

nclack, line 517:
```suggestion
    """Compile @npe2.implements contributions to generate a manifest."""
```

Struggling a bit with the help text here.  Maybe this is ok?

nclack (commented):
I think once the help text for `npe2 compile` is updated this is gtg.  Works well in my hands!

npe2/cli.py
tlambert03: ohhhhh.  yep I'm so sorry. I missed that before!  thanks a lot for catching that!

tlambert03: apologies!  i missed the crux of this comment.  thanks for the correction

tlambert03:
thanks for the review @nclack!  much appreciated.  Since this one isn't very discoverable or "consumer facing" just yet, and since it mostly works,  I'm going to merge.  we can polish off any rough edges before creating docs pages for it

PR #188, opened 2022-06-21
Allow `@implements` in comments, and prevent runtime-arg checking by default

tlambert03:
This PR adds two things to the `npe2.implements` decorators
 
1. Adds a safety measure with a new `ensure_args_valid = False` parameter to the decorators:

```
@npe2.implements.reader(
    id="url_reader",
    title="URL Reader",
    filename_patterns=["http://*", "https://*"],
    ensure_args_valid=True,
)
def url_reader(path: str):
    ...
```

by default, it is `False` ... meaning the decorators do essentially nothing at runtime.  This is a safety measure in case some plugin uses the decorator, and then eventually goes out of spec with the npe2 schema.  If npe2 is updated in the environment, the decorator will just do nothing.  Those errors can still be detected at compile time (when building the manifest), but they should _not_ error at runtime unless the plugin specifically requests it.

It also allows decorators **in comments** to be detected, meaning a plugin can use `npe2 compile` to build a manifest with **no** runtime impact or dependency on npe2 at all.  This should help to address comments about depending on npe2 in @nclack's https://github.com/napari/npe2/pull/75#pullrequestreview-1012855747
It looks like this:

```python
# No import needed!
# @npe2.implements.writer(
#     id="my_single_writer",
#     title="My single-layer Writer",
#     filename_extensions=["*.xyz"],
#     layer_types=["labels"],
# )
def writer_function_single(path: str, layer_data: Any, meta: Dict) -> List[str]:
    ...
```

the only restriction at the moment is that the decorator name _must_ use the fully qualified `npe2.implements.something`, rather than `@implements.something` or `@something`

### DIFF
--- a/npe2/implements.py
+++ b/npe2/implements.py
@@ -222,6 +222,51 @@ class PluginModuleVisitor(ast.NodeVisitor):
         return f"{self.module_name}:{obj_name}"
 
 
+def _parse_comment_call(lines: Sequence[str]) -> Tuple[str, Dict[str, Any]]:
+    """Parse a comment block into a function name and a dictionary of kwargs.
+
+    Parameters
+    ----------
+    lines : Sequence[str]
+        Lines of strings, the coment block.
+
+    Returns
+    -------
+    Tuple[str, Dict[str, Any]]
+        `(name_of_function, kwargs)`
+
+    Examples
+    --------
+    >>> lines = [
+    ...     '# @npe2.implements.writer(',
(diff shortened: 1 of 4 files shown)


tlambert03:
sorry for the hurry on this one guys,  I'm gonna merge this and release, since I have a few other things that are waiting on this release. It should have been in #75 in the first place... and i don't want to release without this in the first implementation of the decorators.  We can of course iterate later if anyone has post-merge concerns

PR #192, opened 2022-06-22
Add `npe2 list` command to discover/display all currently installed plugins

tlambert03:
work in progress, but here's the idea:

Screen Shot 2022-06-21 at 9 08 15 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 **Implemented enhancements:**
 
+- Prevent runtime-arg checking on npe2.implements decorators by default [\#188](https://github.com/napari/npe2/pull/188) ([tlambert03](https://github.com/tlambert03))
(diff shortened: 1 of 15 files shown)


nclack:
I find myself using this constantly bwiw

tlambert03:
> I find myself using this constantly fwiw.

what do you mean?  Like, you check out this branch and run it? 😂 if so we need to get this in for you :)

> will need `rich` in the `install_requires`?

Yep, will do thanks.

tlambert03:
Ok, need to add tests and more docs, but i think the implementation is roughly where I'd have it, so marking ready for first review.  I added an optional `--format` flag. 

Screen Shot 2022-06-24 at 8 02 23 AM

Screen Shot 2022-06-24 at 8 02 42 AM

Screen Shot 2022-06-24 at 8 02 54 AM

tlambert03:
I've merged in #197 here, so that should be reviewed first.  I also added docs and cleaned this up a bit.  So the only thing left now is to complete test coverge

nclack (approved):
Thanks.  Looks great.

PR #193, opened 2022-06-22
Remove typing-extensions imports

tlambert03:
We had some `from typing_extensions` imports, but didn't have it listed in our dependencies.  It's something that ends up being in almost every environment, so we didn't have any issues yet (I guess), but it should have been listed in deps.  Turns out though, that everything we were importing there is now available in python >=3.8, so this PR just changes those imports to `from typing`


nclack (approved):
lgtm

PR #195, opened 2022-06-23
Add docs clarifying menus `when` and `group`

tlambert03:
just clarifying the difference between menu `when` and command `enablement`, along with the `group` field in menus

### DIFF
--- a/npe2/manifest/contributions/_menus.py
+++ b/npe2/manifest/contributions/_menus.py
@@ -10,9 +10,20 @@ class _MenuItem(BaseModel):
     """Generic menu item contribution."""
 
     when: Optional[str] = Field(
-        description="Condition which must be true to show this item"
+        description="Condition which must be true to *show* this item in the menu. "
+        "Note that ``when`` clauses apply to menus and ``enablement`` clauses to "
+        "commands. The ``enablement`` applies to all menus and even keybindings while "
+        "the ``when`` only applies to a single menu."
     )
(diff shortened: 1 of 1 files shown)


nclack:
I do love clarifying docs but I have a question:

Would it be crazy to use `when` for both? Or is there a behavioral difference? Like: when `enablement` is false we might do something like grey out a menu item, but if the `when` clause is false we might not show the menu item.

tlambert03:
> Like: when enablement is false we might do something like grey out a menu item, but if the when clause is false we might not show the menu item.

this is exactly the case.  enablement 

> Would it be crazy to use `when` for both? 

do you mean use the _word_ "when" for both?  Or _only_ have `when` in one of the two fields, and use it to glean the meaning for the other?  
If the former, I'm slightly ambivalent... I think it's kinda nice to have a different word ("enablement" implies that it's always there, just not always enabled, while I can convince myself that "when" means it should only show up in the menu "when). 
If the latter, then I do definitely think we need a field in both `command` and `menu` (the former to indicate when the command could be triggered *at all*, regardless of whether it appears in a menu or not, and the latter indicating *when* it should appear in a menu).  We already have such a distinction in `_layer_actions.py` ... so we do kinda need it

tlambert03:
I'm not sure if your comment was a request for changes, or just a side though.  I hope you don't mind if I merge this, since it's strictly just better documentation of what the current pattern is.  If you want to dig into `when` vs `enablement`, or consider changes to the actual pattern, let's do that in another issue or PR

PR #203, opened 2022-06-27
Fix `npe2 list` when a dotted field key is empty

tlambert03:
Fixes a bug when a dotted field name has been requested in `npe2 list` (such as `package_metadata.version`) if one of the upper level names doesn't exist (i.e. if `package_metadata` is `None`)


nclack (approved):
thanks, I'm going to go ahead and merge

PR #208, opened 2022-07-05
Fix documentation links

melissawm:
This PR fixes several broken links in the documentation, mostly those referring to local anchors that are not well parsed by myst and need their own definitions.

This also fixes one instance where the anchor name (`contributions-sample_data`) was not correctly written from the template.


tlambert03 (approved):
thanks @melissawm!  looks good to me.  still in draft for a reason?

melissawm:
Hi @tlambert03, I was hoping we could get https://github.com/napari/napari/pull/4787 in first, because there have been some subtle changes there. Thanks for checking!

melissawm:
This should be good to go now, thanks!

PR #210, opened 2022-07-18
add back commented command fields

tlambert03:
I need these fields now that were commented out in #49 

cc @nclack

### DIFF
--- a/npe2/manifest/contributions/_commands.py
+++ b/npe2/manifest/contributions/_commands.py
@@ -63,22 +63,21 @@ class CommandContribution(BaseModel):
 
     short_title: Optional[str] = Field(
         None,
-        description="(Optional) Short title by which the command is represented in "
+        description="Short title by which the command is represented in "
         "the UI. Menus pick either `title` or `short_title` depending on the context "
         "in which they show commands.",
     )
     category: Optional[str] = Field(
         None,
-        description="(Optional) Category string by which the command may be grouped "
-        "in the UI",
+        description="Category string by which the command may be grouped in the UI.",
     )
     icon: Optional[Union[str, Icon]] = Field(
         None,
-        description="(Optional) Icon used to represent this command in the UI, e.g. on "
-        "buttons or in menus. These may be superqt fonticon keys, such as "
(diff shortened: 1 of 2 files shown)


npe2/manifest/contributions/_commands.py
```diff
@@ -60,31 +61,32 @@ class CommandContribution(BaseModel):
...
+    )
+    category: Optional[str] = Field(
+        None,
+        description="Category string by which the command may be grouped in the UI.",
+    )
+    icon: Optional[Union[str, Icon]] = Field(
+        None,
+        description="Icon used to represent this command in the UI, on "
+        "buttons or in menus. These may be [superqt](https://github.com/napari/superqt)"
+        " fonticon keys, such as `'fa6s.arrow_down'`; though not that plugins are "
```
nclack, line 78:
```suggestion
        " fonticon keys, such as `'fa6s.arrow_down'`; though note that plugins are "
```

tlambert03: :joy:  that's a very different message indeed

npe2/manifest/contributions/_commands.py
```diff
@@ -60,31 +61,32 @@ class CommandContribution(BaseModel):
...
+        "expected to depend on any fonticon libraries they use, e.g "
+        "[fonticon-fontawesome6](https://github.com/tlambert03/fonticon-fontawesome6).",
+    )
+    enablement: Optional[str] = Field(
+        None,
+        description=(
+            "Condition which must be true to enable the command in the UI "
```
nclack, line 85:
```suggestion
            "Expression which must evaluate as true to enable the command in the UI "
```

Adding an example to the description might be helpful here too just to show precisely what's expected.

nclack (approved):
Looks good!

npe2/manifest/contributions/_commands.py
tlambert03: I agree... but I think we need to make a full page of docs on expressions with all of the valid names that they can use.  It will look like this: https://code.visualstudio.com/api/references/when-clause-contexts

PR #218, opened 2022-07-30
build: remove magicgui dependency

tlambert03:
this removes magicgui from dependencies.  We barely use it (just one isinstance check) and the call to autogeneration which isn't ever called internally).

This guards those two places with a try import...  The `isinstance` check works exactly the same way.  The widget autogeneration will just fail with an informative message if magicgui isn't installed (so, it's up to napari to depend on magicgui rather than npe2)


DragaDoncila (approved):
Looks good

PR #220, opened 2022-07-31
Improve error message on schema validation

Czaki:
When renaming the package I use `_` instead of `-` in the new name and error message does not allow me to understand where is the source of the problem. 

In this PR I improve the error message to have information where is a problem.


tlambert03 (approved):
looks good thanks.  while we're here, I actually think the phrase "Manifest name" that we previously used is a little ambiguous: it could be interpreted as the name of the manifest file, yeah?  What do you think of 'manifest.name' must match package name or "the name field in the manifest must match the name of the package"?

Czaki:
> the name field in the manifest must match the name of the package

I like this, but with provided name or information that `_` are replaced with `-`.

tlambert03:
> > the name field in the manifest must match the name of the package
> 
> I like this, but with provided name or information that `_` are replaced with `-`.

yeah, leave in the bit you added

npe2/manifest/schema.py
```diff
@@ -321,7 +321,9 @@ def _from_entrypoint(
                     meta = PackageMetadata.from_dist_metadata(distribution.metadata)
                     mf.package_metadata = meta
 
-                    assert mf.name == meta.name, "Manifest name must match package name"
+                    assert (
+                        mf.name == meta.name
+                    ), f"Manifest name ({mf.name}) must match package name ({meta.name})"
```
tlambert03, line 326:
```suggestion
                    assert (
                        mf.name == meta.name
                    ), f"The name field in the manifest ({mf.name!r}) must match package name ({meta.name!r})"
```

Issue #54, opened 2021-12-10
Idea: add an npe2 cli tool to replace the cookiecutter.


nclack:
### Description

cookiecutter is pretty limited. It's a pretty limited cli frontend in front of a jinja templating engine.

We could think about creating our own front-end - maybe keeping the jinja template for parts but using the npe2 machinery for manifest creation in particular.  It seems super nice to me to have an `npe2 create` command. The npe2 command becomes something like a one-stop-shop for related tooling.

Also, we could leverage the type information in the manifest to automate some of the ui.  ui could be a cli or a gui.

There was some discussion around this in the 2021 Dec 9 community meeting.

@goanpeca @tlambert03 @sofroniewn

PR #225, opened 2022-08-03
[wip] use copier for npe2 create

tlambert03:
Been playing with using [copier](https://copier.readthedocs.io/en/stable/) as a replacement for cookiecutter.  Generally much happier with it, mostly because of support for 1) conditionals in the questions, 2) much better updating when the upstream template changes 3) validation (after [my PR merges](https://github.com/copier-org/copier/pull/719))

I built a copier template here: https://github.com/tlambert03/napari-plugin-template

this PR would address #54 and add an `npe2 create` command that would bootstrap a new plugin, as @nclack proposed.  Most of the work will be in tailoring that plugin template repo as we want it.  and I also think we should create an actual webpage documentation for it.

if we eventually wanted to go even farther than copier, we could just keep the command

### DIFF
--- a/npe2/cli.py
+++ b/npe2/cli.py
@@ -522,5 +522,34 @@ def compile(
     _pprint_formatted(manifest_string, format)
 
 
+@app.command()
+def create(plugin_name: str):
+    """Create a new napari plugin"""
+    import subprocess
+    from shutil import which
+
+    if not which("copier"):
+        text = typer.style(
+            "This command requires the `copier` package.\n"
+            "Install it now into your current environment? [y/N]",
+            fg="yellow",
+        )
+        if not typer.prompt(text, type=bool, default=False, show_default=False):
+            raise typer.Exit("Aborted")
+
+        copier = "git+https://github.com/tlambert03/copier.git@validation"
+        subprocess.check_call([sys.executable, "-m", "pip", "install", copier])
+
+    subprocess.check_call(
+        [
+            sys.executable,
+            "-m",
+            "copier",
+            "gh:tlambert03/napari-plugin-template",
+            plugin_name,
+        ]
+    )
+
+
 def main():
     app()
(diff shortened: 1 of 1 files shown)


npe2/cli.py
```diff
@@ -522,5 +522,34 @@ def compile(
...
+@app.command()
+def create(plugin_name: str):
+    """Create a new napari plugin"""
+    import subprocess
+    from shutil import which
+
+    if not which("copier"):
+        text = typer.style(
+            "This command requires the `copier` package.\n"
+            "Install it now into your current environment? [y/N]",
```
nclack, line 534:
```suggestion
            "Install it now into your current environment with pip? [y/N]",
```

nclack (approved):
Running through the interactive prompts myself, there were usability improvements but there are a few rough edges too.  Those don't need to be addressed here, but I was wondering if you also felt there were rough edges.

Just being able to do `npe2 create my-plugin` feels so so much better to me than having to look up the cookiecutter template. We could even think about having different kinds of templates ('bare minimum' vs 'everything but the kitchen sink')

There's a lot I don't really like about cookiecutter, so overall this feels like a really nice improvement.

tlambert03:
> there were usability improvements but there are a few rough edges too.

I agree, did you write them down? 😅

nclack:
> > there were usability improvements but there are a few rough edges too.
> 
> I agree, did you write them down? 😅

I did :) Off the top of my head:

* For string defaults like the author field, had to delete each char and then enter the non-default test manually.
* The installed paths reported at the end didn't include the name of the source root. For example, it would say "created 'src/napari.yaml'" instead of "created my-silly-plugin/src/napari.yml".

tlambert03:
> had to delete each char and then enter the non-default test manually.

nclack:
> I might leave this PR as WIP for a bit until that template is ready/better

up to you.  I think this is gtg and the template can get iterated on later.

Issue #226, opened 2022-08-05
Support for plugins using dot notation to indicate namespaces


kmdouglass:
# Summary

Should Napari provide support for plugins whose names use namespaces in dot notation to indicate ownership?

This issue is intended to consolidate discussion on this topic.

# Discussion

PyPI and the Python packaging infrastructure support namespaces to identify ownership. For example, [gp.bootstrap](https://pypi.org/project/gp.bootstrap/) is a package owned by Gael Pasgrimaud, who has multiple PyPI packages starting with the `gp.` namespace.

It appears that Napari currently does not support plugins whose names contain dots to separate namespaces. The following Napari manifest for a fictitious plugin called `kyle-douglass.my-plugin` fails validation: 

```
name: kyle-douglass.my-plugin
contributions:
  commands:
    - id: kyle-douglass.my-plugin.frame_rate_widget
      title: open my widget
      python_name: kyle_douglass.my_plugin._widget:Frame_rate_Widget
    - id: kyle-douglass.my-plugin.Time_scroller_widget
      title: add time scroller widget
      python_name: kyle_douglass.my_plugin._widget:Time_scroller_widget
  widgets:
    - command: kyle-douglass.my-plugin.frame_rate_widget 
      display_name: Plot frame rate
    - command: kyle-douglass.my-plugin.Time_scroller_widget
      display_name: Add time scroller
```

The output of `npe2 validate` is:

```console
Invalid! 2 validation errors for PluginManifest
name
  'kyle-douglass.my-plugin' is not a valid python package name. (type=value_error)
__root__
  Commands identifiers must start with the current package name None
            the following commands where found to break this assumption:
                ['kyle-douglass.my-plugin.frame_rate_widget', 'kyle-douglass.my-plugin.Time_scroller_widget']
 (type=value_error)
```

# Misc.

- Namespaces to identify ownership on PyPI are discussed in [PEP-423](https://peps.python.org/pep-0423/).
- It's possible that Napari plugins and their corresponding PyPI packages need not be the same, but this violates some suggestions in PEP-423 for consistency in package naming.
- Namespace packages are a slightly different concept than namespaces on PyPI and, I think, are unrelated: https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages

tlambert03:
[...]

PR #230, opened 2022-08-08
fix command id validation when dot in package name

tlambert03:
fixes and adds a test for https://github.com/napari/npe2/issues/226#issuecomment-1208095028


tlambert03:
napari test error is unrelated, and due to this: https://github.com/gristlabs/asttokens/issues/85

PR #235, opened 2022-08-13
feat: add icon to manifest

tlambert03:
adds an icon field to the manifest, which will allow us to associate an icon with each plugin in the plugin installer


nclack (approved):
I like it 👍 I imagine emoji support can be added later 🌼

PR #237, opened 2022-08-13
bug: ignore extra fields on manifest

### LINKED ISSUE #236: remove `extra = Extra.forbid` in schema config?
@nclack  ... shouldn't we have `extra = Extra.ignore` in our PluginManifest config?  doesn't having extra=forbid mean that if a plugin wants to add a new field to support a new feature (like visibility, or icon) then they are necessarily making themselves incompatible with older versions of npe2?

added in https://github.com/napari/npe2/pull/14

tlambert03:
closes #236 and reverts a change from #14

While forbidding extra keys on makes for nicer validation if a plugin developer enters a field name incorrectly, it makes it so that we can't add fields to the manifest without making them backwards incompatible.  Normally, adding a field should be an easy migration in a schema, but currently it's hard. 

I'm not quite sure what we should do now to allow plugin developers to adopt new manifest keys like (`icon` or `visibility`) without making their plugins incompatible with any versions of npe2 earlier than our next release.  but this PR will help prevent that going forward.

### DIFF
--- a/npe2/manifest/schema.py
+++ b/npe2/manifest/schema.py
@@ -35,7 +35,7 @@ class DiscoverResults(NamedTuple):
 class PluginManifest(ImportExportModel):
     class Config:
         underscore_attrs_are_private = True
-        extra = Extra.forbid
+        extra = Extra.ignore
         validate_assignment = True
 
     # VS Code uses <publisher>.<name> as a unique ID for the extension
--- a/npe2/manifest/schema.py
+++ b/npe2/manifest/schema.py
@@ -196,17 +196,20 @@ class PluginManifest(ImportExportModel):
 
     @root_validator
     def _validate_root(cls, values: dict) -> dict:
+        mf_name = values.get("name")
+
         # validate schema version
         declared_version = Version.parse(values.get("schema_version", ""))
         current_version = Version.parse(SCHEMA_VERSION)
         if current_version < declared_version:
-            raise ValueError(
-                f"The declared schema version '{declared_version}' is "
-                f"newer than npe2's schema version: '{current_version}'. You may "
-                "need to upgrade npe2."
+            import warnings
+
+            warnings.warn(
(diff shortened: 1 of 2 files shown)


nclack (approved):
Thanks. I think this is the right approach.

"Ignore" rather than "forbid" opens the possibility of folks having spelling errors in their manifest that get silently ignored/defaulted. But I think compatibility is more important.

With respect to the new fields, we could bump the schema version so old versions will indicate the user should upgrade their npe2...not much different than getting an error on the field but maybe more helpful.

tlambert03:
> we could bump the schema version so old versions will indicate the user should upgrade their npe2...not much different than getting an error on the field but maybe more helpful.

yeah, I like this thanks.  I forgot we had this line in there: great! 😅 

https://github.com/napari/npe2/blob/3ead8db38029ff6ef0389b450ff417c1d50dc82d/npe2/manifest/schema.py#L202-L206

tlambert03:
however...  now I'm kinda wishing we made that a warning and not an exception.

I've bumped the SCHEMA_VERSION here to 0.2.0 and made the "version too high" a warning that says "things may break, you should update npe2".  Since, it's not _guaranteed_ to break, we might as well let it try?

nclack:
I wish there a way to do `extra = warn`. 

Can we flip on forbid during `npe2 validate`? maybe with a flag? The possible spelling errors are huanting me. If it's remotely possible, I can turn that into an issue. Shouldn't block this pr.

tlambert03:
Pydantic doesn't offer it, but I'm sure we could rig something in both cases (we can do anything! 🤓)

Yeah, make new issue, npe2 validate could generally be made to do good stuff

tlambert03:
well, this auto commit is at least comforting, that every `0.1.0` was able to just become `0.2.0` :joy: https://github.com/napari/npe2api/commit/043879b554e0d63d1c5242cb726348202f8d4861

PR #242, opened 2022-08-16
fix: relax display name validation

tlambert03:
The display name validation is too strict, both in terms of length and characters allowed.  It should allow things like colons, periods, etc...  This relaxes it to a string from 3-90 characters, and only restricts the possible values of the first character


npe2/manifest/_validators.py
```diff
@@ -60,8 +60,7 @@ def python_name(name: str) -> str:
 def display_name(v: str) -> str:
     if not DISPLAY_NAME_PATTERN.match(v):
         raise ValueError(
-            f"{v} is not a valid display_name.  The display_name must "
-            "be 3-40 characters long, containing printable word characters, "
+            f"{v} is not a valid display_name. It must be 3-40 characters long, "
```
nclack, line 63:
```suggestion
            f"{v} is not a valid display_name. It must be 3-90 characters long, "
```

nclack (approved):
Looks fine!

npe2/manifest/_validators.py
tlambert03: lol.  I corrected that like 3 times and _still_ ended up with 40 :joy:  thank you

PR #245, opened 2022-09-12
Allow for lists within the command chain

mstabrin:
This change allows to proper visualize cli information if the nesting contains lists.


npe2/cli.py
```diff
@@ -231,7 +246,7 @@ def _make_rows(pm_dict: dict, normed_fields: Sequence[str]) -> Iterator[List]:
 @app.command()
 def list(
     fields: str = typer.Option(
-        "name,version,npe2,contributions",
+        "name,version,npe2,contributions,contributions.widgets.display_name",
```
tlambert03, line 249: i don't think the default value should be changed here.  too specific to a particular use case.  If you want to get all the widget display names, you should use the `fields` option of `npe2 list` manually

mstabrin: Yes indeed, I reverted the change :)

src/npe2/cli.py
```diff
@@ -105,6 +105,9 @@ def _pprint_table(
             val = ""
             if isinstance(r, dict):
                 val = ", ".join(f"{k} ({v})" for k, v in r.items())
+            # list if overridden by function name
+            elif isinstance(r, type([])):
+                val = ", ".join(str(_) for _ in r)
```
Czaki, line 110:
```suggestion
            elif isinstance(r, list):
                val = ", ".join(str(text) for text in r)
```

`_` means in python unused variable

src/npe2/cli.py
```diff
@@ -212,9 +215,21 @@ def _make_rows(pm_dict: dict, normed_fields: Sequence[str]) -> Iterator[List]:
...
-                while parts and hasattr(val, "__getitem__"):
-                    val = val[parts.pop(0)]
+                val = [info]
+                while parts:
+                    part = parts.pop(0)
+                    new_val = []
+                    for cur_val in val:
+                        if hasattr(cur_val, "__getitem__"):
+                            # list if overridden by function name
+                            if isinstance(cur_val[part], type([])):
```
Czaki, line 225:
```suggestion
                            if isinstance(cur_val[part], list):
```

src/npe2/cli.py
```diff
@@ -212,9 +215,21 @@ def _make_rows(pm_dict: dict, normed_fields: Sequence[str]) -> Iterator[List]:
...
+                                new_val.extend(cur_val[part])
+                            else:
+                                new_val.append(cur_val[part])
+                    val = new_val
```
Czaki, line 229: This does not look like reproduce and extend of the previous logic. It may break tests.

jni:
@mstabrin do you think you might have a chance to pick this up again soon? 🙏

nclack:
I'm going to go ahead and close this since there hasn't been activity for a while. Can always reopen if there's more interest/activity.

PR #248, opened 2022-10-07
Add count of discovered plugins

Czaki:
Add methods to better trace changes in plugins to have the possibility to message users.
For example to, inform users that there is a need to restart napari after installing the npe2 plugin.

### DIFF
--- a/npe2/_plugin_manager.py
+++ b/npe2/_plugin_manager.py
@@ -272,7 +272,7 @@ class PluginManager:
         
         Returns
         -------
-        dicover_count : int
+        discover_count : int
             Number of discovered plugins
             
         """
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -71,6 +71,7 @@ testing =
     numpy
     pytest
     pytest-cov
+    jsonschema
```
nclack, line 74: nice catch

nclack (approved):
thanks! looks good to me

setup.cfg
Czaki: Result of setup a fresh environment.

psobolewskiPhD:
Works well with https://github.com/napari/napari/pull/5190/files, really important fix, because without the warning to the user it seems like the plugin failed to install.

PR #256, opened 2022-11-24
Fix PackageMetadata validation error with extra provided field

aganders3:
The `PackageMetadata` class is configured to ignore extra fields but instantiation can still fail (raises a `KeyError`) when actually provided an extra field due to this custom root validator. Based on the config this does not seem like the intended behavior.

```
class PackageMetadata(BaseModel):
    ....
    class Config:
        extra = Extra.ignore
    ...
    @root_validator(pre=True)
    def _validate_root(cls, values):
        if "metadata_version" not in values:
            fields = cls.__fields__
            mins = {fields[n].field_info.extra.get("min_ver", "1.0") for n in values}
            values["metadata_version"] = str(max(float(x) for x in mins))
        return values
```

This PR just changes that validator skip over any values not in the list of expected fields during this step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
           python -m pip install -e ./napari-from-github
           python -m pip install -e .[json]
           # bare minimum required to test napari/plugins
-          python -m pip install pytest scikit-image[data] zarr xarray hypothesis matplotlib pyside6
+          python -m pip install pytest scikit-image[data] zarr xarray hypothesis matplotlib
 
       - name: Run napari plugin headless tests
         run: pytest -W 'ignore::DeprecationWarning' napari/plugins napari/settings napari/layers napari/components
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
           python -m pip install -e ./napari-from-github
           python -m pip install -e .[json]
           # bare minimum required to test napari/plugins
-          python -m pip install pytest scikit-image[data] zarr xarray hypothesis matplotlib
+          python -m pip install pytest scikit-image[data] zarr xarray hypothesis matplotlib pyside6
 
       - name: Run napari plugin headless tests
         run: pytest -W 'ignore::DeprecationWarning' napari/plugins napari/settings napari/layers napari/components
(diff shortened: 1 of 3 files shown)


nclack (approved):
Thanks! 

You can remove the pyside dependency in the ci. It'll cause a test to fail, but that should get fixed in #255.

.github/workflows/ci.yml
```diff
@@ -65,7 +65,7 @@ jobs:
           python -m pip install -e ./napari-from-github
           python -m pip install -e .[json]
           # bare minimum required to test napari/plugins
-          python -m pip install pytest scikit-image[data] zarr xarray hypothesis matplotlib
+          python -m pip install pytest scikit-image[data] zarr xarray hypothesis matplotlib pyside6
```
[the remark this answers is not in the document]
nclack: to clarify, the `CI/napari tests` check should fail, but that's ok.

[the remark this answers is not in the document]
aganders3: Done - thanks!

goanpeca (approved):

tests/test_package_meta.py
```diff
@@ -17,3 +17,24 @@ def test_package_metadata_version():
...
 def test_hashable():
     hash(PackageMetadata(name="test", version="1.0"))
+
+
+def test_package_metadata_extra_field():
+    pkg = {
+        "name": "test",
+        "version": "1.0",
+        "maintainer": "bob",
+        "extra_field_that_is_definitely_not_in_the_model": False,
```
goanpeca, line 27: Love this key name 😄

tests/test_package_meta.py
```diff
@@ -17,3 +17,24 @@ def test_package_metadata_version():
...
+    }
+
+    try:
+        p = PackageMetadata(**pkg)
```
goanpeca, line 31: @aganders3 I am confused, so with the fix it should not fail with extra fields? So why is the test checking that it fails. Or am I missing something 🤔

aganders3: You're right - the try/except is a little ugly but is basically rewriting whatever error would come up into an `AssertionError` for consistency with pytest expectations.

Then after parsing succeeds the other assertions check known fields are present/correct and the extra field is ignored/dropped. Does that make sense?

goanpeca: Yep :)

PR #261, opened 2022-12-14
Expose `syntax_style`

brisvag:
While working on fixing napari themes (see [this zulip conversation](https://napari.zulipchat.com/#narrow/stream/212875-general/topic/theme.20not.20respected.20in.20qss/near/315705427)) I noticed that some fields that are available in napari are not exposed/tested here: `icon` was unused, and `syntax_style` was not exposed. I added them to the specs and description.

Also, I noticed that if you pass wrong/nonexisting keys, `npe2` does not complain. Is this intended?

### DIFF
--- a/_docs/example_manifest.yaml
+++ b/_docs/example_manifest.yaml
@@ -42,6 +42,7 @@ contributions:
     - label: "Monokai"
       id: "monokai"
       type: "dark"
+      syntax_style: "monokai"
       colors:
         canvas: "#000000"
         console: "#000000"
@@ -51,6 +52,7 @@ contributions:
         secondary: "#f8f8f2"
         highlight: "#e6db74"
         text: "#a1ef34"
+        icon: "#a1ef34"
         warning: "#f92672"
         current: "#66d9ef"
   sample_data:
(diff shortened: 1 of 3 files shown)


tlambert03:
> if you pass wrong/nonexisting keys, npe2 does not complain. Is this intended?

Yes, otherwise every new field we add to the schema becomes a breaking change for anyone on an older version of npe2

brisvag:
Makes sense. It would be nice if the `npe2 validate` command could tell you that you have keys in your schema that do not exist in the current `npe2` version. Would help with debugging!

brisvag:
I can't tell if this failure is a real bug. Are we missing `napari-svg` for macos?

tlambert03:
just an internet problem

brisvag:
Still failing :/

tlambert03:
still think it's internet... see this test: https://github.com/napari/npe2/actions/runs/3693708290/jobs/6256673381  where everything passed but codecov

tlambert03:
see https://github.com/napari/npe2/pull/255

nclack (approved):
lgtm! Just updated to fix the tests. If they all pass now, I'll look to merge tomorrow.

Issue #262, opened 2022-12-15
Evaluate alternative pypi api for tests


nclack:
I'm not sure - there's a bit of history here including past requests from napari: https://github.com/pypi/warehouse/issues/9156. Their timeline is very vague but I would suspect they will not remove it until there is an equivalent new API. If not I would hope for a concrete deprecation period to be announced at some point.

One idea is to use the XML RPC API and fall back on the current method. The XML RPC API has some methods already deprecated that now raise a `RuntimeError` so we can probably anticipate the same path. I'm likewise not too confident in the stability of the interface we're currently scraping, but there doesn't seem to be a clear path forward for this capability.

Regardless I think it could be saved for another PR - for now I guess I would just recommend adding the change proposed in my other comment above https://github.com/napari/npe2/pull/255#issuecomment-1332683255. It's small and could be added here or I can make another PR for that.

_Originally posted by @aganders3 in https://github.com/napari/npe2/issues/255#issuecomment-1349561325_

aganders3:
I did some more evaluation here and implemented a method to fetch via XML-RPC and fallback to scraping, but in testing I found that the data returned by the XML-RPC API are not quite the same as the current implementation.

The main difference is that the XML-RPC API will return _more_ plugins. The scraping method seems to only find packages whose latest version includes the `Framework :: napari` classifier. For example [`napari-webcam`](https://github.com/haesleinhuepf/napari-webcam) removed the classifier in v0.1.19, and is not found via scraping. Meanwhile the XML-RPC API will still return the v0.1.18 that had the classifier.

Right now there are to be 16 such packages:
```
E         {'flood-napari': '0.0.1',
E          'napari-czifile': '0.0.2',
E          'napari-imagecodecs': '0.0.2',
E          'napari-lfdfiles': '0.0.2',
E          'napari-mahotas-image-processing': '0.1.2',
E          'napari-manual-split-and-merge-labels': '0.1.3',
E          'napari-netpbmfile': '0.0.2',
E          'napari-oclrfc': '0.4.5',
E          'napari-oiffile': '0.0.2',
E          'napari-sdtfile': '0.0.2',
E          'napari-sim-SIMulator': '0.0.1',
E          'napari-tifffile': '0.0.2',
E          'napari-webcam': '0.1.18',
E          'nucleaizer-backend': '0.2.0',
E          'ome-zarr': '0.0.19',
E          'pyclesperanto-prototype': '0.10.9'}
```

[...]

PR #265, opened 2023-01-05
Update fetch API to improve reliability

aganders3:
This draft includes two proposed changes:
1. Fetch available plugins via the PyPI XML-RPC API, which is generally much faster and seems more reliable than scraping, but is considered legacy and planned to be deprecated.
2. Fallback on scraping as we do now, but with improved reliability by explicitly setting the ordering to "Date last updated" which results in more stable results.

See #262 for more discussion on whether the first is desirable - I will leave this as draft until that issue is resolved, but I think the second change should be incorporated either way.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,6 +26,7 @@ install_requires =
     PyYAML
     appdirs
     build
+    packaging
     psygnal>=0.3.0
     pydantic
     pytomlpp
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -26,6 +26,7 @@ install_requires =
     PyYAML
     appdirs
     build
+    packaging
```
aganders3, line 29: N.B. this is already a transitive dependency I think

PR #270, opened 2023-01-27
Fix link to magicgui objects.inv in intersphinx

melissawm:
Fixes one broken link in the napari documentation.

In `_docs/templates/_npe2_contributions.md.jinja` (which is used to generate https://napari.org/stable/plugins/contributions.html, the lack of correct URL for the magicgui docs inventory file will create a broken link to `magicgui.widgets.Widget` (see https://napari.org/stable/plugins/contributions.html#contributions-widgets)


nclack (approved):
Thanks!

PR #272, opened 2023-02-27
docs: fix private magicgui import in docs build

tlambert03:
cc @Czaki re https://github.com/napari/napari/pull/5592#issuecomment-1447169318

### DIFF
--- a/_docs/render.py
+++ b/_docs/render.py
@@ -6,7 +6,7 @@ import os
 import re
 import sys
 from contextlib import contextmanager
-from functools import lru_cache
+from functools import lru_cache, partial
 from inspect import getsource
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


_docs/render.py
```diff
@@ -103,7 +102,7 @@ def _build_example(contrib: Executable) -> str:
 
     if not callable(func):
         return ""
-    if isinstance(func, MagicFactory):
+    if isinstance(func, partial):
```
Czaki, line 105: Is this true for any partial or `hasattr` check should be added?

tlambert03: true for any partial

Czaki:
There is a need to bump isort version as poetry introduces backward compatibility breaking change in some releases.

Issue #271, opened 2023-02-14
`filename_patterns` for readers should be case insensitive


DragaDoncila:
### Description

Currently our filename matching for readers is case sensitive (as evidenced below). I did a quick Google on what's good practice but I couldn't find a silver bullet "they should/shouldn't be" case sensitive. This current behaviour, however, is incompatible with `napari`'s behaviour which as of [#5121](https://github.com/napari/napari/pull/5121) lowercases filenames. 

If we agree that the patterns should be case insensitive, we should update the `iter_compatible_readers` check and the documentation.

Otherwise we should revert napari #5121? I don't think we can do one and not the other.
 
```python
In [20]: from npe2 import PluginManager, DynamicPlugin

In [21]: pm = PluginManager()

In [22]: reader_plugin = DynamicPlugin(name='tif_reader', plugin_manager=pm)

In [23]: @reader_plugin.contribute.reader(filename_patterns=["*.TIF"], accepts_directories=False)
    ...: def get_reader(path):
    ...:     def read(path):
    ...:         return [(None,)]
    ...:     return read
    ...: 

In [24]: pm.register(reader_plugin.manifest)

In [25]: list(pm.iter_compatible_readers("myfile.TIF"))
Out[25]: [ReaderContribution(command='tif_reader.get_reader', filename_patterns=['*.TIF'], accepts_directories=False)]

In [26]: list(pm.iter_compatible_readers("myfile.tif"))
Out[26]: []

```

nclack:
Looks like there's some support for case-insensitive in this [zulip thread](https://napari.zulipchat.com/#narrow/stream/309872-plugins/topic/case.20insensitive.20pattern.20matching.3F).  I also like it - it's the default behavior I think folks will expect.

There's a risk that some reader might _need_ case-sensitive. But we can add a flag to the `ReaderContribution` to support that use case when it crops up.

Does anyone know of case-sensitive use cases?

psobolewskiPhD:
Note that that napari PR only lower cases when checking for plugins, so it's the worst of both worlds 🤣 
See: https://github.com/napari/napari/issues/5663#issuecomment-1481886108
I don't think anyone would expect a different plugin for .jpg and .JPG so I think making everything case insensitive is the way to go.
So the question is just where is the best place to add the `.lower()`...

nclack:
Looks like we have consensus on the _should_ so I changed the title of the issue (I hope that was alright).

PR #275, opened 2023-03-24
Bugfix: use .lower() to make paths & pattern fnmatch case  insensitive

psobolewskiPhD:
First shot at making the matching of paths to the declared extensions case insensitive.
Closes: https://github.com/napari/napari/issues/5663
Closes https://github.com/napari/npe2/issues/271

@DragaDoncila I'm not sure this is how you had it in mind?
I'm not super familiar with the code base so not sure I'm approaching it the right way vs. just bandaid bugfix.

I added two tests and make fixes to get the tests to pass.
Maybe this is too kludgey...

### DIFF
--- a/tests/test__io_utils.py
+++ b/tests/test__io_utils.py
@@ -26,14 +26,6 @@ def test_read_with_unknown_plugin(uses_sample_plugin):
         read(["some.fzzy"], plugin_name="nope", stack=False)
 
 
-def test_read_uppercase_extension(uses_sample_plugin):
-    # sample plugin hard-codes lower case and returns this error
-    # so the error ensures the matching was case-insensitive
-    # and that the sample plugin received full path (with case)
-    with pytest.raises(ValueError, match="Test plugin should"):
-        read(["some.FZZY"], stack=False)
-
-
 def test_read_with_no_plugin():
     # no plugin passed and none registered
     with pytest.raises(ValueError, match="No readers returned"):
@@ -44,16 +36,6 @@ def test_read_return_reader(uses_sample_plugin):
     data, reader = read_get_reader("some.fzzy")
     assert data == [(None,)]
     assert reader.command == f"{SAMPLE_PLUGIN_NAME}.some_reader"
-
-
-def test_read_uppercase_extension_return_reader(uses_sample_plugin):
-    # sample plugin hard-codes lower case and returns this error
-    # so the error ensures the matching was case-insensitive
-    # and that the sample plugin received full path (with case)
-    with pytest.raises(ValueError, match="Test plugin should"):
-        data, reader = read_get_reader("some.FZZY")
-        assert data == [(None,)]
-        assert reader.command == f"{SAMPLE_PLUGIN_NAME}.some_reader"
 
 
 def test_read_return_reader_with_stack(uses_sample_plugin):
@@ -112,3 +94,91 @@ def test_read_non_global_pm():
         return _reader
 
     assert io_utils._read(["some.fzzy"], stack=False, _pm=pm) == [(None,)]
+
+
+def test_read_uppercase_extension():
+    pm = PluginManager()
+    plugin = DynamicPlugin("tif-plugin", plugin_manager=pm)
+    path = "something.TIF"
+
+    # reader should be compatible despite lowercase pattern
+    @plugin.contribute.reader(filename_patterns=["*.tif"])
+    def get_read(path):
+        if path.lower() != path:
+            # if this error is raised we can be certain path is unchanged
+            raise ValueError("Given path contains capitals.")
+
+        def read(paths):
+            return [(None,)]
+
+        return read
+
+    with pytest.raises(ValueError, match="Given path contains capitals."):
+        io_utils._read([path], stack=False, _pm=pm)
+
+
+@pytest.mark.parametrize(
+    "path", ["some_zarr_directory.ZARR", "some_zarr_directory.Zarr"]
+)
+def test_read_zarr_variants(path: str):
+    pm = PluginManager()
+    plugin = DynamicPlugin("zarr-plugin", plugin_manager=pm)
+
+    # reader should be compatible despite lowercase pattern
+    @plugin.contribute.reader(filename_patterns=["*.zarr"])
+    def get_read(path):
+        if path.lower() != path:
+            # if this error is raised we can be certain path is unchanged
+            raise ValueError("Given path contains capitals.")
+
+        def read(paths):
+            return [(None,)]
+
+        return read
+
+    with pytest.raises(ValueError, match="Given path contains capitals."):
+        io_utils._read([path], stack=False, _pm=pm)
+
+
+@pytest.mark.parametrize(
+    "path", ["some_two_ext_file.TAR.gz", "some_two_ext_file.TAR.GZ"]
+)
+def test_read_tar_gz_variants(path: str):
+    pm = PluginManager()
+    plugin = DynamicPlugin("tar-gz-plugin", plugin_manager=pm)
+
+    # reader should be compatible despite lowercase pattern
+    @plugin.contribute.reader(filename_patterns=["*.tar.gz"])
+    def get_read(path):
+        if path.lower() != path:
+            # if this error is raised we can be certain path is unchanged
+            raise ValueError("Given path contains capitals.")
+
+        def read(paths):
+            return [(None,)]
+
+        return read
+
+    with pytest.raises(ValueError, match="Given path contains capitals."):
+        io_utils._read([path], stack=False, _pm=pm)
+
+
+@pytest.mark.parametrize("path", ["some_directory.Final", "some_directory.FINAL"])
+def test_read_directory_variants(path: str):
+    pm = PluginManager()
+    plugin = DynamicPlugin("directory-plugin", plugin_manager=pm)
+
+    # reader should be compatible despite lowercase pattern
+    @plugin.contribute.reader(filename_patterns=["*.final"], accepts_directories=True)
+    def get_read(path):
+        if path.lower() != path:
+            # if this error is raised we can be certain path is unchanged
+            raise ValueError("Given path contains capitals.")
+
+        def read(paths):
+            return [(None,)]
+
+        return read
+
+    with pytest.raises(ValueError, match="Given path contains capitals."):
+        io_utils._read([path], stack=False, _pm=pm)
--- a/tests/test__io_utils.py
+++ b/tests/test__io_utils.py
@@ -185,4 +185,4 @@ def test_read_directory_variants(path: str, tmp_path: Path):
         return read
 
     with pytest.raises(ValueError, match="Given path contains capitals."):
-        io_utils._read([path], stack=False, _pm=pm)
+        io_utils._read([str(new_path)], stack=False, _pm=pm)
--- a/tests/test__io_utils.py
+++ b/tests/test__io_utils.py
@@ -166,7 +166,9 @@ def test_read_tar_gz_variants(path: str):
 
 
 @pytest.mark.parametrize("path", ["some_directory.Final", "some_directory.FINAL"])
-def test_read_directory_variants(path: str):
+def test_read_directory_variants(path: str, tmp_path: Path):
+    new_dir = tmp_path / path 
+    new_dir.mkdir()
     pm = PluginManager()
     plugin = DynamicPlugin("directory-plugin", plugin_manager=pm)
 
--- a/tests/test__io_utils.py
+++ b/tests/test__io_utils.py
@@ -16,6 +16,19 @@ def test_read_with_unknown_plugin(uses_sample_plugin):
     # no such plugin name.... skips over the sample plugin & error is specific
     with pytest.raises(ValueError, match="Plugin 'nope' was selected"):
         read(["some.fzzy"], plugin_name="nope", stack=False)
+        
+def test_read_uppercase_extension(uses_sample_plugin):
+    # sample plugin hard-codes lower case and returns this error
+    # so the error ensures the matching was case-insensitive
+    # and that the sample plugin received full path (with case)
+    with pytest.raises(ValueError, match="Test plugin should"):
+        read(["some.FZZY"], stack=False)
+
+
+def test_read_with_plugin(uses_sample_plugin):
+    # no such plugin name.... but skips over the sample plugin
+    with pytest.raises(ValueError):
+        read(["some.fzzy"], plugin_name="nope", stack=False)
 
 
 def test_read_with_no_plugin():
@@ -28,6 +41,16 @@ def test_read_return_reader(uses_sample_plugin):
     data, reader = read_get_reader("some.fzzy")
     assert data == [(None,)]
     assert reader.command == f"{SAMPLE_PLUGIN_NAME}.some_reader"
+
+
+def test_read_uppercase_extension_return_reader(uses_sample_plugin):
+    # sample plugin hard-codes lower case and returns this error
+    # so the error ensures the matching was case-insensitive
+    # and that the sample plugin received full path (with case)
+    with pytest.raises(ValueError, match="Test plugin should"):
+        data, reader = read_get_reader("some.FZZY")
+        assert data == [(None,)]
+        assert reader.command == f"{SAMPLE_PLUGIN_NAME}.some_reader"
 
 
 def test_read_return_reader_with_stack(uses_sample_plugin):
--- a/tests/test__io_utils.py
+++ b/tests/test__io_utils.py
@@ -16,7 +16,8 @@ def test_read_with_unknown_plugin(uses_sample_plugin):
     # no such plugin name.... skips over the sample plugin & error is specific
     with pytest.raises(ValueError, match="Plugin 'nope' was selected"):
         read(["some.fzzy"], plugin_name="nope", stack=False)
-        
+
+
 def test_read_uppercase_extension(uses_sample_plugin):
     # sample plugin hard-codes lower case and returns this error
     # so the error ensures the matching was case-insensitive
--- a/tests/test__io_utils.py
+++ b/tests/test__io_utils.py
@@ -26,12 +26,6 @@ def test_read_uppercase_extension(uses_sample_plugin):
         read(["some.FZZY"], stack=False)
 
 
-def test_read_with_plugin(uses_sample_plugin):
-    # no such plugin name.... but skips over the sample plugin
-    with pytest.raises(ValueError):
-        read(["some.fzzy"], plugin_name="nope", stack=False)
-
-
 def test_read_with_no_plugin():
     # no plugin passed and none registered
     with pytest.raises(ValueError, match="No readers returned"):
--- a/tests/test__io_utils.py
+++ b/tests/test__io_utils.py
@@ -125,7 +125,7 @@ def test_read_zarr_variants(path: str):
     plugin = DynamicPlugin("zarr-plugin", plugin_manager=pm)
 
     # reader should be compatible despite lowercase pattern
-    @plugin.contribute.reader(filename_patterns=["*.zarr"])
+    @plugin.contribute.reader(filename_patterns=["*.zarr"], accepts_directories=True)
     def get_read(path):
         if path.lower() != path:
             # if this error is raised we can be certain path is unchanged
--- a/tests/test__io_utils.py
+++ b/tests/test__io_utils.py
@@ -121,7 +121,7 @@ def test_read_uppercase_extension():
     "path", ["some_zarr_directory.ZARR", "some_zarr_directory.Zarr"]
 )
 def test_read_zarr_variants(path: str, tmp_path: Path):
-    new_dir = tmp_path / path 
+    new_dir = tmp_path / path
     new_dir.mkdir()
     pm = PluginManager()
     plugin = DynamicPlugin("zarr-plugin", plugin_manager=pm)
@@ -167,7 +167,7 @@ def test_read_tar_gz_variants(path: str):
 
 @pytest.mark.parametrize("path", ["some_directory.Final", "some_directory.FINAL"])
 def test_read_directory_variants(path: str, tmp_path: Path):
-    new_dir = tmp_path / path 
+    new_dir = tmp_path / path
     new_dir.mkdir()
     pm = PluginManager()
     plugin = DynamicPlugin("directory-plugin", plugin_manager=pm)
--- a/tests/test__io_utils.py
+++ b/tests/test__io_utils.py
@@ -173,7 +173,7 @@ def test_read_directory_variants(path: str, tmp_path: Path):
     plugin = DynamicPlugin("directory-plugin", plugin_manager=pm)
 
     # reader should be compatible despite lowercase pattern
-    @plugin.contribute.reader(filename_patterns=["*.final"], accepts_directories=True)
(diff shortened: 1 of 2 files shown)


npe2/_plugin_manager.py
```diff
@@ -135,14 +135,20 @@ def iter_compatible_readers(self, paths: List[str]) -> Iterator[ReaderContributi
             return
         assert isinstance(path, str)
 
+        # use lower() to make matching case-insensitive
+        path = path.lower()
+
         if os.path.isdir(path):
```
Czaki, line 141: on UNIX systems (MacOS, Linux) filesystem is cases sensitive:

`/Users/` will be converted to `/users/` which most probably does not exists.

npe2/_plugin_manager.py
```diff
@@ -135,14 +135,20 @@ def iter_compatible_readers(self, paths: List[str]) -> Iterator[ReaderContributi
...
             yield from (r for pattern, r in self._readers if pattern == "")
         else:
             # not sure about the set logic as it won't be lazy anymore,
             # but would we yield duplicate anymore.
             # above does not have have the unseen check either.
             # it's easy to make an iterable version if we wish, or use more-itertools.
-            yield from {r for pattern, r in self._readers if fnmatch(path, pattern)}
+            # match against pattern.lower() to make matching case insensitive
+            yield from {
+                r for pattern, r in self._readers if fnmatch(path, pattern.lower())
```
Czaki, line 150: fnmatch should be case insensitive
https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatch

It looks like problem is somewhere else.

Czaki (commented):
I do not have time for full review, but it looks like the solution is not proper.

npe2/_plugin_manager.py
psobolewskiPhD: Only on windows? See the link:
>Both parameters are case-normalized using [os.path.normcase()](https://docs.python.org/3/library/os.path.html#os.path.normcase).

https://docs.python.org/3/library/os.path.html#os.path.normcase
which says
> On other operating systems, return the path unchanged.

npe2/_plugin_manager.py
psobolewskiPhD: Hmm, i'm on macOS and it still matches correctly. 
But yeah, probably safer to use something like:
```
base, ext = os.path.splitext(path)
path = os.path.join(base + ext.lower())
```

psobolewskiPhD: both `os.path.isdir` and os.path.splitext` work on `.lower()` even though the real paths on my computer have cases.

psobolewskiPhD: Can instead just move the `.lower()` to the `fnmatch` here:
https://github.com/psobolewskiPhD/npe2/blob/b6cf2cf8d72ff287051b1e02cd79f9a8e537ec0f/npe2/_plugin_manager.py#L150

psobolewskiPhD: OK, so this actually does nothing, because `paths` is set to lower in `io_utils.py` `_read` (see below).
So some duplication on my part. Not sure where it's more logical to do this.
`io_utils.py`: `_read` or `read_get_reader`
or here in `_plugin_manager` (deleting the changes in `io_utils`)

npe2/_plugin_manager.py
psobolewskiPhD: @Czaki 
Here's some tests on macOS:
```
In [9]: test_path = Path('/Users/piotrsobolewski/Downloads/IMG_2767.JPG')

In [10]: test_path.exists()
Out[10]: True

In [11]: fnmatch.fnmatch(test_path, '*.jpg')
Out[11]: False

In [12]: fnmatch.fnmatch(test_path, '*.JPG')
Out[12]: True

In [13]: fnmatch.fnmatch(test_path.lower(), '*.jpg')
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[13], line 1
----> 1 fnmatch.fnmatch(test_path.lower(), '*.jpg')

AttributeError: 'PosixPath' object has no attribute 'lower'

In [14]: fnmatch.fnmatch(str(test_path).lower(), '*.jpg')
Out[14]: True

```

npe2/_plugin_manager.py
psobolewskiPhD: Ok, i avoid the duplication now. `lower()` is used in `_read` only, I think this makes more sense.

psobolewskiPhD: OK, that was annoying because _read can take a str or a list.
I put the check back in _plugin_manager, in the correct _iter_compatible_readers
(sorry made a mess of the commits!)

npe2/_plugin_manager.py
```diff
@@ -142,7 +142,10 @@ def iter_compatible_readers(self, paths: List[str]) -> Iterator[ReaderContributi
...
```
Czaki, line 147:
```suggestion
                r for pattern, r in self._readers if fnmatch(path, pattern.lower()) or fnmatch(path, pattern.upper())
```

psobolewskiPhD: huh, instead of doing .lower() on paths in iter_compatible_readers?
clever.

But what if there's mixed case?

Czaki: Maybe https://docs.python.org/3/library/fnmatch.html#fnmatch.translate and run `re.match with` https://docs.python.org/3/library/re.html#re.IGNORECASE ?

psobolewskiPhD: seems really complex, is it better than just lower casing for the comparison?

Czaki: you mean `fnmatch(path.lower(), pattern.lower())`?

psobolewskiPhD: yea. it's the simplest and I think it should work...

psobolewskiPhD:
Test fails are same as this PR: https://github.com/napari/npe2/pull/276
I think it may be related to napari-ndtiffs being updated?
https://github.com/tlambert03/napari-ndtiffs
Will try to test some tomorrow

Czaki:
> I think it may be related to napari-ndtiffs being updated?

Yes. I think It is related.

The best will be create separate PR with fix. 

@tlambert03 what you think will be best. Pin `napari-ndtiffs` to older release or remove this test?

tlambert03:
Ah, hah!
Sorry about that, who'd have thought updating a package to npe2 would break npe tests 😂 (I know I know, my own damn fault, it was a shortcut for testing the npe1 adapter).  I don't have a preference for how it's fixed.  Both pinning or deleting the test sound like a good idea... a fixture is probably the "best" but more laborious.  So maybe just pin

npe2/_plugin_manager.py
psobolewskiPhD: Ok, using `path.lower(), pattern.lower()` in the fnmatch works in napari, but not in the existing tests i mimicked, because, as far as I can tell, the sample_plugin hard-codes lower case extension in the reader function:
https://github.com/napari/npe2/blob/d7329b7088f59ac743a67ff47aca32067b5a246a/tests/sample/my_plugin/__init__.py#L24-L41

My tests error with `E           ValueError: Test plugin should not receive unknown data` so I think this is actually the expected result of everything working properly.
By using `.lower()` only in the `fnmatch` the plugin gets the real path and in this case, errors. On the other hand, making the actual path have `ext.lower()` as before, would make sample_plugin handle this case and the new tests would pass.
So options are to assert something else, go back to .lower() on the path, or use .lower() in sample_plugin to make the actual plugin also case insensitive.

I think I lean towards option 1 or option 3. npe2 should be case-insensitive, but if a plugin is actually case sensitive for some reason, then it should be allowed to raise an error?

DragaDoncila: Hmm I don't think we can/should distinguish between path being case insensitive and pattern being case insensitive. I think our reading in general should be either case sensitive or case insensitive. I think the same should be true of plugins (so, plugins should be case insensitive), but the `get_reader` functions complicate things a little, since they may be checking things about the path that are not true if we try to enforce case insensitivity by passing `path.lower()`.

I think we should make paths fully case insensitive for reading so:

- match `pattern.lower()` against `ext.lower()`
- we always pass through the unchanged path because I think that makes most sense
- add docs to contribution guide for readers to advise that pattern matching will be case insensitive
    - and say that we encourage readers to also be case insensitive
    - but if you desperately need case sensitivity you could check in your `get_reader` function (but I don't think we even mention this tbh)
- we update the cookiecutter reader to recommend case insensitivity

I had a look through all the `filename_patterns` being declared. Only 5 out of 92 readers declare any pattern with capital letters in it:

```
napari-annotatorj: ['<EDIT_ME>']
napari-deepfinder: ['*.mrc', '*.map', '*.rec', '*.h5', '*.tif', '*.TIF', '*.xml', '*.ods', '*.xls', '*.xlsx']
napari-pdr-reader: ['*.fits', '*.FITS', '*.lbl', '*.img', '*.LBL', '*.IMG']
napari-rioxarray: ['*.vrt', '*.tif', '*.tiff', '*.TIF', '*.TIFF', '*.img', '*.lbl', '*.cub', '*.fits', '*.IMG', '*.LBL', '*.CUB', '*.FITS']
napari-tomocube-data-viewer: ['*.TCF']
```

The first is clearly a placeholder. `napari-deepfinder` explicitly makes sure it's case insensitive for TIFs, `napari-pdr-reader` also makes its patterns case insensitive, and so does `napari-rioxarray` for all but one of the extensions - but it then goes through and l[owercases the path](https://github.com/AndrewAnnex/napari-rioxarray/blob/main/src/napari_rioxarray/_reader.py#L42) as it comes in anyway. I couldn't find the repo for `napari-tomocube-data-viewer` so not sure if it truly is case sensitive or not. Anyway, given the above, I would say we adopt case insensitivity for both patterns and paths, and update docs accordingly.

psobolewskiPhD:
OK, I may have totally botched this with trying to reconcile the `src` change in conflicts...

...but I think I have it now how we discussed:
1. ext.lower() is used for comparisons to the pattern
2. but the actual path is passed to the plugin
So as a result I made the tests expect the ValueError from sample plugin.

tests/test__io_utils.py
```diff
@@ -30,6 +38,16 @@ def test_read_return_reader(uses_sample_plugin):
     assert reader.command == f"{SAMPLE_PLUGIN_NAME}.some_reader"
 
 
+def test_read_uppercase_extension_return_reader(uses_sample_plugin):
```
DragaDoncila, line 41: Ok now that I look at it I think this is kind of a misleading way to check for case sensitivity. I think we should update the sample plugin to be case insensitive, or only pass it lowercase paths - it's not great to co-opt the exception to test case sensitivity. In this PR we should instead update [this test](https://github.com/napari/npe2/blob/14b2c055865001a60b7e15af90bd38771c75d41a/tests/test_contributions.py#L62) to check case insensitivity (or maybe just add a new test using that pattern).

This test also tries to make sure that we pass the full path, but it doesn't really achieve this, since it's `iter_compatible_readers` that **might** change it, and that's not called here. Testing for this should occur on the `napari` side. So I think here we should actually just remove this test

DragaDoncila (commented):
This looks good @psobolewskiPhD, and I checked that it works! I left a comment about making the tests more explicit. Also I think we should update the contribution docs [here](https://github.com/napari/npe2/blob/eb6703e52c0354ef9204c2c706bea8885cfc6e5b/npe2/manifest/readers.py#L13) to mention the case insensitivity - I'm pretty sure these get built into the reference on napari.org

tests/test__io_utils.py
psobolewskiPhD: I did it this way to ensure that the matching was insensitive, but that the real path was passed so a plugin could be case sensitive internally. I thought this was what we wanted from the community meeting 😅

tests/test__io_utils.py
```diff
@@ -30,6 +38,16 @@ def test_read_return_reader(uses_sample_plugin):
...
+    # sample plugin hard-codes lower case and returns this error
+    # so the error ensures the matching was case-insensitive
+    # and that the sample plugin received full path (with case)
+    with pytest.raises(ValueError, match="Test plugin should"):
```
Czaki, line 45: Where is this exception raised? I see three lines; this construction should only be used for a single line. Based on the code, it looks like lines 47 and 48 are not executed.

Czaki (changes requested):
We should create at least the following tests and get working code for this:

1) `some_directory_name.FINAL`, `some_directory_name.Final` - should return directory reader
2) `some_zarr_directory.zarr`, `some_zarr_directory.ZARR`, `some_zarr_directory.Zarr` should return zarr reader (test may need to implement dummy reader). 
3) `some_two_ext_file.tar.gz`, `some_two_ext_file.TAR.gz`, `some_two_ext_file.TAR.GZ`, etc

The current solution will not catch properly any of these tests. 
The only tests that are added to this PR do not test these cases and will not prevent future break of this use case

tests/test__io_utils.py
DragaDoncila: Oh wait, I got confused heh 😅 . So yes, we do definitely want to test that the path casing is preserved when passed to the read function.

I do still think this is a slightly odd way of testing it - i.e. it relies on a side effect of the sample plugin. So my recommendation for testing that the path is unchanged is that we use a dynamic plugin - I just tried doing that though and it didn't work so... please hold hehe

DragaDoncila: Ok @psobolewskiPhD after #285 is merged, you should be able to write a test similar to this:

```ipython
In [2]: from npe2.io_utils import _read

In [3]: from npe2 import PluginManager, DynamicPlugin

In [4]: pm = PluginManager()

In [5]: plugin = DynamicPlugin('tif-read', plugin_manager=pm)

In [6]:     @plugin.contribute.reader
   ...:     def read_path(path):
   ...:         if path.lower() != path:
   ...:             raise ValueError("Path has capitals in it as we expected!")
   ...:         def _reader(path):
   ...:             return [(None,)]
   ...: 
   ...:         return _reader
   ...: 
 # we now expect this to raise a ValueError because the path has capitals in it, so you could use `with pytest.raises()` here
In [7]: _read(["some.BLAH"], stack=False, _pm=pm, plugin_name='tif-read')
```

Of course you don't **have** to raise an error, you could return a read function that returned some kind of known layer data that you then check against. But an error raised is easy to check for. Lmk if this makes sense!

psobolewskiPhD: @DragaDoncila I'm not sure I totally get it. This creates a reader, but one that has no filename pattern right? So the changes implemented in this PR won't be tested I dont think and this test will always raise the error, so always pass?
Maybe I should edit the sample plugin to check the path that is passed (like it does) but not error and just report back that it was upper or lower?

tests/test__io_utils.py
psobolewskiPhD: Yes, sorry copy paste error. I will delete those, but looks like the whole test will be different anyways.

tests/test__io_utils.py
DragaDoncila: Sorry @psobolewskiPhD you can provide filename patterns in the reader contribution's decorator - I was messing around with other things and forgot to add the filename pattern back sorry. If you search the code base for `DynamicPlugin` you should see some examples. I can post the full test later, I'm just on my phone. This approach is also how we should test the other suggestions from Grzegorz.

DragaDoncila:
```ipython
In [1]: from npe2 import PluginManager, DynamicPlugin

In [2]: pm = PluginManager()

In [3]: reader_plugin = DynamicPlugin(name='tif_reader', plugin_manager=pm)

In [4]: @reader_plugin.contribute.reader(filename_patterns=["*.tif"], accepts_directories=False)
   ...: def get_read(pths):
   ...:     # we passed a path with capitals, and it should stil contain them
   ...:     if pths.lower() == pths:
   ...:         raise ValueError("Path was changed before being passed to reader")
   ...:     def read(paths):
   ...:         return [(None,)]
   ...:     return read
   ...: 
```

psobolewskiPhD: OK, that makes more sense and it's what I sort of dug down into reading the code, so I guess I'm on the right track. But when I do put filename_patterns there i keep getting
`ValueError: No readers returned data for ['some.JPG']` and `ValueError: No readers returned data for ['some.jpg']` regardless of the pattern, so I must still be doing something wrong. I'll take another look tomorrow.

DragaDoncila: Hmmm interesting. Let me know if you wanna pair on this, otherwise I can also take a look tomorrow!

psobolewskiPhD: We can pair, could be fun.
Alternately, maybe just implementing this in the sample_plugin would be more robust anyways? Just need to modify the manifest to include the other permitted patterns and tweak the code to not error on the upper case extension as it does now. Could then probably parameterize just the paths to test.

DragaDoncila: @psobolewskiPhD this test passes locally for me and seems to do everything we need. I like using `DynamicPlugin` because it's very flexible. We can use the same pattern of test to try directories, capital pattern and lowercase path, etc. If we update the `sample_plugin` it's a lot more annoying to rig up all the different options.

```python
def test_read_uppercase_extension():
    pm = PluginManager()
    plugin = DynamicPlugin("tif-plugin", plugin_manager=pm)
    path = "something.TIF"

    # reader should be compatible despite lowercase pattern
    @plugin.contribute.reader(filename_patterns=["*.tif"])    
    def get_read(path):
        if path.lower() != path:
            # if this error is raised we can be certain path is unchanged
            raise ValueError("Given path contains capitals.")
        def read(paths):
            return [(None,)]
        return read

    with pytest.raises(ValueError, match="Given path contains capitals."):
        io_utils._read([path], stack=False, _pm=pm)
```

psobolewskiPhD:
I re-did the tests per @DragaDoncila and @Czaki  which revealed that double extensions were not handled (only the last). So now I switched to use `pathlib.Path.suffixes`.
Further, URI were also not handled, so I check using `urllib.parse.urlparse(path).scheme` first, before modifying the path.
 
I think this covers everything?

Czaki:
Why .zarr test do not contain `accepts_directories` when the final contains?

psobolewskiPhD:
> Why .zarr test do not contain accepts_directories when the final contains?

Because I forgot? 😅 
Fixed, I think.

DragaDoncila (approved):
Thanks @psobolewskiPhD this looks good to me!

tests/test__io_utils.py
```diff
@@ -94,3 +94,91 @@ def _reader(path):
...
+            # if this error is raised we can be certain path is unchanged
+            raise ValueError("Given path contains capitals.")
+
+        def read(paths):
+            return [(None,)]
+
+        return read
+
+    with pytest.raises(ValueError, match="Given path contains capitals."):
+        io_utils._read([path], stack=False, _pm=pm)
```
Czaki, line 184: Currently, this test is similar to Zarr one. My idea when I was requesting it was to check if a plain directory reader would handle a directory with an "extension" 

I'm not sure if it should be  `filename_patterns=["*"]` or lack of this parameter 

```suggestion
@pytest.mark.parametrize("path", ["some_directory.Final", "some_directory.FINAL"])
def test_read_directory_variants_no_ext(path: str):
    pm = PluginManager()
    plugin = DynamicPlugin("directory-plugin", plugin_manager=pm)

    # reader should be compatible despite lowercase pattern
    @plugin.contribute.reader(filename_patterns=["*"], accepts_directories=True)
    def get_read(path):
        if path.lower() != path:
            # if this error is raised we can be certain path is unchanged
            raise ValueError("Given path contains capitals.")

        def read(paths):
            return [(None,)]

        return read

    with pytest.raises(ValueError, match="Given path contains capitals."):
        io_utils._read([path], stack=False, _pm=pm)
```

psobolewskiPhD: Oh, sorry I didn't get that.
Looking at the code for dirs:
https://github.com/napari/npe2/blob/00d2bc227437a37f276c1ba9fad108a263504469/src/npe2/_plugin_manager.py#L138-L139
So I think in theory the extension shouldn't matter?

But I'm not sure these tests are checking directories effectively? Because `os.path.isdir(path)` will return false in every case.

tests/test__io_utils.py
```diff
@@ -94,3 +94,91 @@ def _reader(path):
...
+
+
+@pytest.mark.parametrize(
+    "path", ["some_zarr_directory.ZARR", "some_zarr_directory.Zarr"]
+)
+def test_read_zarr_variants(path: str):
```
Czaki, line 123:
```suggestion
def test_read_zarr_variants(path: str, tmp_path: Path):
    new_dir = tmp_path / path 
    new_dir.mkdir()
```

tests/test__io_utils.py
```diff
@@ -94,3 +94,91 @@ def _reader(path):
...
+            # if this error is raised we can be certain path is unchanged
+            raise ValueError("Given path contains capitals.")
+
+        def read(paths):
+            return [(None,)]
+
+        return read
+
+    with pytest.raises(ValueError, match="Given path contains capitals."):
+        io_utils._read([path], stack=False, _pm=pm)
```
Czaki, line 140:
```suggestion
        io_utils._read([str(new_dir)], stack=False, _pm=pm)
```

tests/test__io_utils.py
```diff
@@ -94,3 +94,91 @@ def _reader(path):
...
+
+
+@pytest.mark.parametrize("path", ["some_directory.Final", "some_directory.FINAL"])
+def test_read_directory_variants(path: str):
```
Czaki, line 167:
```suggestion
def test_read_directory_variants(path: str, tmp_path: Path):
    new_dir = tmp_path / path 
    new_dir.mkdir()
```

tests/test__io_utils.py
Czaki, line 184:
```suggestion
        io_utils._read([str(new_path)], stack=False, _pm=pm)
```

tests/test__io_utils.py
```diff
@@ -94,3 +94,91 @@ def _reader(path):
...
+    pm = PluginManager()
+    plugin = DynamicPlugin("directory-plugin", plugin_manager=pm)
+
+    # reader should be compatible despite lowercase pattern
+    @plugin.contribute.reader(filename_patterns=["*.final"], accepts_directories=True)
```
Czaki, line 172:
```suggestion
    @plugin.contribute.reader(filename_patterns=["*"], accepts_directories=True)
```

Czaki (commented):
The test is not only to validate the current implementation but mainly to prevent introducing regression in another commit. 

I add a suggestion to create dummy dirs in test.

Maybe we should also create dummy tiff and tar.gz files

tests/test__io_utils.py
```diff
@@ -96,14 +97,17 @@ def _reader(path):
...
     plugin = DynamicPlugin("tif-plugin", plugin_manager=pm)
+
     path = "something.TIF"
+    mock_file = tmp_path / path
+    mock_file.touch()
 
     # reader should be compatible despite lowercase pattern
     @plugin.contribute.reader(filename_patterns=["*.tif"])
-    def get_read(path):
+    def get_read(path=mock_file):
```
psobolewskiPhD: Is this the sort of thing you had in mind @Czaki ?

psobolewskiPhD: used the same below for tar.gz

Czaki (approved):
looks great

Issue #280, opened 2023-04-06
convert to hatch and src layout?


tlambert03:
this issue https://github.com/pypa/setuptools/issues/3557 is really making cross-package editable installs difficult... meaning it's hard to run `pip install -e npe2` in a napari dev environment and have it work... you get something like this:
```
In [1]: import npe2

In [2]: npe2
Out[2]: <module 'npe2' (<_frozen_importlib_external._NamespaceLoader object at 0x10cfb3f40>)>
```

supposedly, moving to a src layout fixes that for setuptools.   but hatch just gets this right regardless, and I'd love to convert this repo over to hatch.  any objections @DragaDoncila @nclack @jni @Czaki ?

Czaki:
I do not have big experience with hatch. Especially with the short trial, I do not be able to enforce running tests via hatch. 

But independently of the select build backend, I support switching to `src` layout.

In the context of hatch I prefer to either switch to more than one project or keep everything on setuptools as it is better known by core-devs. But maybe I'm wrong and more core-devs are familiar with hatch

Czaki:
> I do not be able to enforce running tests via hatch." ... but no need to push it if you're uncomfortable with it.

In theory, hatch could replace `tox` as it supports creating venv and running tests in it. But I was not able to get this working (also do not spend to much time on it). 


> I'm curious though, have you, however, tried to install package A (that uses setuptools) in editable mode and then tried to use package A from local package B? I

I do not remember such case but I may have outdated setuptools in part of the environments.

tlambert03:
> In theory, hatch could replace tox as it supports creating venv and running tests in it.

yeah, but I'm not talking about that here (besides, we don't even use tox in npe2).  I was just talking about using `hatchling` as a build backend

tlambert03:
anyway, see #281  for the src layout ... would be trivial to move to a hatchling build backend from there

nclack:
It seems like converting to a src layout is the important thing. 💕 src layout.

Hatch looks like a tool that makes that easier to setup and maintain.  I like that it seems to help w version updates/publishing workflow.  I don't have any experience with it, but it looks good. Seems worth a try.

tlambert03:
to be clear, all this really entails is changing the `build-system` table in pyproject to 

```toml
[build-system]
requires = ["hatchling", "hatch-vcs"]
build-backend = "hatchling.build"

[tool.hatch.version]
source = "vcs"
```

Czaki:
[...]

PR #281, opened 2023-04-06
refactor: use src layout and pyproject.toml

tlambert03:
see #280  for a little discussion.  setuptools editable installs now do a bad job unless you're using a src layout.  so this is necessary


tlambert03:
the one broken napari test needs to be fixed on the napari side since #276 ... (and testing/fixing that was actually the thing that got me annoyed about setuptools in the first place)

tlambert03:
(see https://github.com/napari/napari/pull/5698 for fix on napari side)

pyproject.toml
```diff
@@ -3,12 +3,90 @@
...
+]
+classifiers = [
+    "Development Status :: 3 - Alpha",
+    "License :: OSI Approved :: BSD License",
+    "Natural Language :: English",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
```
Czaki, line 26: What is blocking us from adding support of python 3.11 for npe2 (I know problems with napari itself).

Czaki (approved):
Two minor comments

pyproject.toml
tlambert03: Nothing, but yes, out of scope here

pyproject.toml
```diff
@@ -3,12 +3,90 @@
...
+    "Typing :: Typed",
+]
+dependencies = [
+    "PyYAML",
+    "appdirs",
+    "build",
```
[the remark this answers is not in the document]
tlambert03: Perhaps, but can't remember, and out of scope here

pyproject.toml
```diff
@@ -29,23 +107,63 @@ line-length = 88
...
+    "I",   # isort
+    "B",   # flake8-bugbear
+    "C4",  # flake8-comprehensions
+    "TID", # flake8-tidy-imports
+    "RUF", # ruff-specific rules
 ]
 
 [tool.ruff.per-file-ignores]
-"npe2/cli.py" = ["B008","A00"]
+"src/npe2/cli.py" = ["B008", "A00"]
```
nclack, line 124: Add comments for these?  Something like:

```python
# B008 - (flake8) Do not perform function call {name} in argument defaults
"src/npe2/cli.py" = ["B008", "A00"]
```

I tried to look up A00 but couldn't find it. Just want to double check you didn't mean something like A001.

nclack (approved):
lgtm!

Left one slightly out-of-scope comment. I had a question about the ruff ignore rules, but that predates this pr.

PR #285, opened 2023-04-13
fix: pass command registry to rdr.exec in io_utils._read

tlambert03:
this fixes an issue wherein `npe2.read` doesn't find a command registered with a non-global PluginManager instance


DragaDoncila (approved):
Yay it works! Thanks @tlambert03

PR #288, opened 2023-05-24
Fix converting of editable package

Czaki:
When I try to convert PartSeg to npe2 I found that there is a problem with the proper finding path to the module when the package is installed in editable mode.

Maybe it is somehow connected to changes to setuptools mentioned by @tlambert03 in the past.


src/npe2/__main__.py
Czaki, line 1: I have added this file to simplify debug my allow run by `python -m`

src/npe2/_inspection/_from_npe1.py
```diff
@@ -517,7 +517,13 @@ def get_top_module_path(package_name, top_module: Optional[str] = None) -> Path:
         top_module = top_mods[0]
 
     path = Path(dist.locate_file(top_module))
-    assert path.is_dir()
+    if not path.is_dir() and dist.files:
+        for f_path in dist.files:
+            if "__editable__" in f_path.name:
+                path = Path(f_path.read_text().strip()) / top_module
+                break
```
jni, line 524: @Czaki I have zero context here so take my review with a grain of salt, but if there are no tests here, could you at least add comments about why these lines are needed? eg "if the given npe1 plugin is installed in editable mode, then X doesn't work so we must do Y"

tlambert03:
Yes @Czaki, this is the weird editable mode used by setuptools for a pyproject only package

Czaki:
@tlambert03 partseg has `setup.py` (concat changelog with readme) , `setup.cfg`, and `pyproject.toml` files. So it looks like it is enough to contain `pyproject.toml`. And it looks like the problem surface on the drop of npe1 may be big without handling this. 

How did you think? Should I implement some simple package to write a test for it, or just add comments?

tlambert03:
Would be great to test it. I can't remember, did I make a test case for a setup.cfg repo here?  If so, could you use it as a template to copy?

Czaki:
@tlambert03 @jni I have finally created the proper test.

tlambert03 (approved):

PR #289, opened 2023-05-31
Change ArrayLike type to use read only properties

andy-sweet:
# Description

The `np.ndarray` type does not seem to satisfy the `ArrayLike` type. This PR fixes that by using read-only properties instead of plain settable attributes.

Here's a simple reproducer for mypy.

```python
import numpy as np
from npe2.types import ArrayLike

a: np.ndarray = np.array([])

def foo(b: ArrayLike) -> None:
    print(b)

foo(a)
```

For me, mypy fails before this change, but passes afterwards. Let me know if there's a way to add some form of this to automated tests/CI.

# Context

I encountered this while writing a plugin where I'm aiming for full(ish) typing coverage. I understand the `ArrayLike` type has other problems with respect to some napari layer data types (which have their own problems), but it seems like it should at least be satisfied by `np.ndarray`. This is also more consistent with [napari's layer data protocol](https://github.com/napari/napari/blob/46dcc9c0976c35ec9dc64be60235f4addcadaa5c/napari/layers/_data_protocols.py#L71).

### DIFF
--- a/src/npe2/types.py
+++ b/src/npe2/types.py
@@ -30,13 +30,16 @@ PythonName = NewType("PythonName", str)
 
 class ArrayLike(Protocol):
     @property
-    def shape(self) -> Tuple[int, ...]: ...
+    def shape(self) -> Tuple[int, ...]:
+        ...
 
     @property
-    def ndim(self) -> int: ...
+    def ndim(self) -> int:
+        ...
 
     @property
-    def dtype(self) -> "np.dtype": ...
+    def dtype(self) -> "np.dtype":
+        ...
 
     def __array__(self) -> "np.ndarray":
         ...  # pragma: no cover


aganders3:
LGTM - I tested your fix with the provided example and confirmed your results. I agree this is the correct way to represent read-only attributes in a Protocol.

kne42:
is there a reason we aren't just using [`numpy.typing.ArrayLike`](https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike)?

andy-sweet:
> is there a reason we aren't just using [`numpy.typing.ArrayLike`](https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike)?

I'd prefer that too.

Looks like there's [some argument against napari's `ArrayLike`](https://github.com/napari/napari/blob/46dcc9c0976c35ec9dc64be60235f4addcadaa5c/napari/types.py#L37), but that may be less applicable for npe2 which is only describing parameter values for initializers.

aganders3:
I'm not too familiar with the history or usage here, but poking around I see that `numpy.typing.ArrayLike` is a bit looser and  accepts things like `list[list[int]]`.

Take this script for example - both functions return data that fail with `add_image`, but mypy only reports an error for `foo`:

```python
import numpy as np
from napari.viewer import Viewer
from npe2.types import ArrayLike


def foo() -> ArrayLike:
    return [[1, 2, 3], [4, 5, 6]]


def bar() -> np.typing.ArrayLike:
    return [[1, 2, 3], [4, 5, 6]]


viewer = Viewer()
try:
    viewer.add_image(foo())
except Exception as e:
    print("failed to add image with `foo()`")
    print(e)

try:
    viewer.add_image(bar())
except Exception as e:
    print("failed to add image with `bar()`")
    print(e)
```

```
❯ python arraylike.py 
failed to add image with `foo()`
'list' object has no attribute 'size'
failed to add image with `bar()`
'list' object has no attribute 'size'

❯ mypy arraylike.py 
arraylike.py:7:12: error: Incompatible return value type (got "List[List[int]]", expected "ArrayLike")  [return-value]
        return [[1, 2, 3], [4, 5, 6]]
               ^~~~~~~~~~~~~~~~~~~~~~
Found 1 error in 1 file (checked 1 source file)
```

tlambert03:
Yeah arraylike is probably too lose.  It's anything that can be _cast_ to an array (including scalars, etc)... here, we're just trying to indicate the protocol of an array object, without restricting to numpy arrays.  Andy your solution is great.  There are ways to test this (https://pypi.org/project/pytest-mypy-plugins/), and I have used it in other projects... but it's a pain, and adds significant maintenance burden to keep working.  So let's not bother

tlambert03:
(In other words, if you want to say something must have a shape and a dtype, then arraylike won't work unless you always call np.asarray inside the body of the function... we're not implying that here with this object)

PR #292, opened 2023-06-05
add graph layer 

JoOkuma:
Hi.

This PR adds "graph" to the list of existing layers.
It's required by PR napari/napari#5861 which adds a new napari graph layer.


jwindhager:
Ah, too bad I only see this now, already had a branch on this (but no PR)... https://github.com/jwindhager/npe2/tree/graph-layer

I'll check whether there are any differences and report back!

jwindhager:
Indeed, there seem to be a few changes missing. I'll contribute them to this PR tomorrow

JoOkuma:
Thanks, @jwindhager , I merged your PR to my branch.

jni (approved):

jni:
@nclack I don't really know this repo 😅 so if you could review+merge that would be great. 😊

src/npe2/_inspection/_from_npe1.py
```diff
@@ -377,6 +377,9 @@ def napari_get_writer(self, impl: HookImplementation):
         )
         return NotImplemented  # pragma: no cover
 
+    def napari_write_graph(self, impl: HookImplementation):
+        self._parse_writer(impl, "graph")
```
nclack, line 381: Is this function required?

I don't recall `napari_plugin_engine` having a `napari_write_graph` callback.

If it _is_ required, it would be nice to have it tested. Adding the hook to the test npe1 plugin should do the trick. See `tests/npe1-plugin/npe1_module/__init__.py` for where to add a `@napari_hook_implementation` stub.

nclack (commented):
I had a question, but once that's resolved this looks pretty good!

src/npe2/_inspection/_from_npe1.py
JoOkuma: @jwindhager could you take a look at this?
This change came from your PR.

jwindhager: Hi @nclack and thanks for pinging me @JoOkuma 

You're right of course, the npe1 obviously doesn't have such a callback and this is therefore dead code. I assume we won't backport support for graph plugins to the npe1? In this case, we can just remove the function.

```suggestion
```

nclack: Thanks @jwindhager!

I went ahead and accepted the suggestion (hope that was ok @JoOkuma). I'm not aware of any plans to backport.  If I'm wrong though, we can always add this back.

nclack (approved):
lgtm!  The test failure is related to #301 and shouldn't block merging imo.  I'll give this 24 hr and merge if there are no objections!

PR #294, opened 2023-06-05
remove tomlpp

### LINKED ISSUE #291: pytomlpp dependecny make it hard to install on aarch64 (e.g docker for mac).
* npe2 version: stable/latest
* Python version: 3.10/3.11
* Operating System: Mac/Linux

### Description

pytomlpp does not have either conda, or wheels for aarch64.

### What I Did

```
$ docker run -i -t continuumio/miniconda3 /bin/bash
$ pip install pytomlpp # or conda-install ... fails.
```

It's possible to install if you 1) create an env that install conda-forge `compilers`, and pip install to build it.

### suggestion

- Use fallback toml library and make pytomlpp optional when possible ?
- See if upstream can build wheels ?

tlambert03:
closes #291 by using tomli instead

### DIFF
--- a/src/npe2/manifest/_bases.py
+++ b/src/npe2/manifest/_bases.py
@@ -3,7 +3,6 @@ from contextlib import contextmanager
 from pathlib import Path
 from typing import Callable, Dict, Optional, Union
 
-import pytomlpp as toml
 import yaml
 from pydantic import BaseModel, PrivateAttr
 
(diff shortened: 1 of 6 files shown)


src/npe2/manifest/_bases.py
```diff
@@ -74,13 +75,18 @@ def from_file(cls, path: Union[Path, str]):
...
+                import tomli as tomllib  # type: ignore [no-redef]
+
+            loader = tomllib.load
         elif path.suffix.lower() in (".yaml", ".yml"):
             loader = yaml.safe_load
         else:
             raise ValueError(f"unrecognized file extension: {path}")  # pragma: no cover
 
-        with open(path, encoding="utf-8") as f:
+        with open(path, mode="rb") as f:
```
Czaki, line 89: why change form text to binary?

tlambert03: because `.load` requires it, and otherwise you get an exception

Czaki: And `yaml.safe_loader` accept both binary and text format? I try to check the documentation but not found. Only `stream` argument without explanation.

Czaki (approved):

PR #301, opened 2023-06-28
Add check for passed reader contribution

DragaDoncila:
Following the report [here](https://napari.zulipchat.com/#narrow/stream/212875-general/topic/reader.20plugin.20issue.20with.200.2E4.2E18) , this PR updates the `_read` function to check if a specific reader contribution of the form `plugin_name.reader_contribution` was passed as `plugin_name`. This worked before #297 (though it was **not** part of the intended public API), but will not work with `npe2>0.7.0` without this change.

Note that this PR does not throw a specific error if the given plugin doesn't exist - we can add a check for that but it's actually non-trivial since the only public reader method atm  is `iter_compatible_readers`. We arguably should never be passing a non-existent plugin this level deep so maybe that's ok.

No doubt `napari` tests will now fail, but I will go and make those updates separately...

cc @Czaki @psobolewskiPhD


DragaDoncila:
Will update this with an additional test and fix pre-commit checks if people are happy with this direction

jni (approved):

jni:
Ok I was a little hasty with my approval, mostly to indicate that I *am* happy with the direction, but in a longer review with @Czaki during community meeting we discussed the following issues:

- The codecov failure was confusing but the [indirect changes tab](https://app.codecov.io/gh/napari/npe2/pull/301/indirect-changes) shows that there is now an unreachable ValueError clause in [lines 194-198](https://github.com/DragaDoncila/npe2/blob/34df97e23c8f06297b2fc61db5f56da2c1acde43/src/npe2/io_utils.py#L194-L198).
- This failure (and my failure to see it) suggests that the nested logic is too complicated now. Ideally, the code should aim to have a flat structure, e.g. `if plugin_name and plugin not in _pm.available_plugins` `...` `elif plugin_name and not contrib:` `...` `elif plugin_name and contrib` etc.
- the plugin "nope" used in testing doesn't exist, but that's not clear from the name: it could also be a test plugin that does exist but never returns a reader, or has no valid contributions. Instead it would be good to change the name to something clearer, like "non-existent-plugin".
- Using `split` without the `maxsplit=` argument is a bit risky because users might include more than one `.` in the plugin+contribution string, even if that's not valid, and this might result in a confusing error message down the line. It's probably safer to use `maxsplit=1` to find the plugin, contrib split.

Just had a chat with @DragaDoncila who says fixes are incoming though! 😊

src/npe2/io_utils.py
```diff
@@ -167,12 +171,98 @@ def _read(
...
+    chosen_compatible_readers = list(
+        filter(lambda rdr: rdr.plugin_name == plugin, compat_readers)
+    )
+    if contrib:
+        chosen_compatible_readers = list(
+            filter(
+                lambda rdr: rdr.command.split(".")[1] == contrib,
+                chosen_compatible_readers,
+            )
+        )
```
aganders3, line 259:
```suggestion
    chosen_compatible_readers = [
        rdr for rdr in compat_readers
        if rdr.plugin_name == plugin
        and (contrib == "" or rdr.command == plugin_name)
    ]
```

aganders3 (commented):
Looks good to me, but I humbly submit some unsolicited nitpicky readability suggestions.

src/npe2/io_utils.py
DragaDoncila: Thank you 🙏

DragaDoncila:
Thank you @aganders3 those were great! So much neater now 😊 I've added more tests and I think the only remaining failure is `napari_tests` but the failure isn't user facing I'm 99.9% sure, but rather just changed logic. I don't know how to make 100% sure though or what the best order of events is.

DragaDoncila:
Idk what this coverage failure is all of a sudden 🥲

aganders3:
Coverage is just failing to upload. I was hoping it was a transient error but maybe a token expired?

src/npe2/io_utils.py
```diff
@@ -167,12 +171,94 @@ def _read(
...
+    # no choice was made and there's no readers to try
+    if not plugin_name and not len(compat_reader_names):
+        raise ValueError(helper_error_message)
+
+    # user didn't make a choice and we have some readers to try, return them
+    if not plugin_name:
+        return compat_readers
+
+    # user made a choice and it exists, but it may not be a compatible reader
+    plugin, _, contrib = plugin_name.partition(".")
```
nclack, line 246: `contrib` isn't used

```suggestion
    plugin, _, _ = plugin_name.partition(".")
```

src/npe2/io_utils.py
```diff
@@ -154,9 +154,13 @@ def _read(
...
         _pm = PluginManager.instance()
 
-    for rdr in _pm.iter_compatible_readers(paths):
-        if plugin_name and rdr.plugin_name != plugin_name:
-            continue
+    # get readers compatible with paths and chosen plugin - raise errors if
+    # choices are invalid or there's nothing to try
+    chosen_compatible_readers = _get_compatible_readers_by_choice(
+        plugin_name, paths, _pm
+    )
```
nclack, line 161: The comment is helpful. Consider adding an assert after line 161 that sort of checks the invariant holds - it's the kind of thing that would save me some time debugging later on if I broke the rule.

```suggestion
    # get readers compatible with paths and chosen plugin - raise errors if
    # choices are invalid or there's nothing to try
    chosen_compatible_readers = _get_compatible_readers_by_choice(
        plugin_name, paths, _pm
    )
    assert chosen_compatible_readers, "Expected an exception before this point"
```

nclack (approved):
LGTM! Left a couple suggestions that you can take or leave.

DragaDoncila:
@nclack I've updated with your suggestions and I think this can go in. Note that #6016 will contain the fix for the currently failing `napari` test. We shouldn't release `npe2` until I've added comprehensive tests in #6016 to make sure we're not breaking a use case with the changes here.

PR #310, opened 2023-08-25
DOC Update and clarify supported widget contributions

lucyleeow:
This PR:

* Clarifies that the only supported widget parent classes are `QtWidgets.QWidget` or `magicgui.widgets.Widget`. The "such as" suggested that other parent classes could be supported
* Removes documented support for wrapper functions - going forward we do not want to support wrapper functions. (I.e. we do not want new plugins to use wrappers). it is difficult to type (both the wrapper, which will just be a normal function, and the return object, which could be many things) and is too flexible and difficult maintain.
    * I think we want to be backwards compatible but this is outside the scope of this PR and should be discussed in: https://github.com/napari/napari/issues/6174 

Context - this was noted as part of the menu contributions work (https://github.com/napari/napari/issues/6174)

### DIFF
--- a/_docs/templates/_npe2_widgets_guide.md.jinja
+++ b/_docs/templates/_npe2_widgets_guide.md.jinja
@@ -17,14 +17,17 @@ specifically requested by the user.
 ```
 
 The widget specification requires that the plugin provide napari with a
-*callable* object that, when called, returns an *instance* of a widget
-(where a "widget" is an instance of `QtWidgets.QWidget` or `magicgui.widgets.Widget`).
+*callable* object that, when called, returns an *instance* of a widget.
+Here "widget" means a subclass of `QtWidgets.QWidget` or `magicgui.widgets.Widget`,
+or a `FunctionGui`. Additionally, the plugin can provide an arbitrary function if using
+'autogenerate', which requests that napari autogenerate a widget using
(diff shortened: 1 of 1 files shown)


lucyleeow:
Note this has been discussed with @Czaki and @DragaDoncila but other opinions welcome and I am happy to change this PR. I will also open a discussion on zulip.

DragaDoncila (approved):
Looks good to me, thanks @lucyleeow!

jni:
TIL what a "wrapper function" is. 😅 Can we include in the merge message the comment on the rationale for removing wrapper support? From napari/napari#6174:

> After discussion with @DragaDoncila and @Czaki, we think we don't want to support 3 because it is difficult to type (both the wrapper, which will just be a normal function, and the return object, which could be many things) and is too flexible and difficult maintain. To keep backwards compatibility, 3 will only be supported in the plugin menu.

lucyleeow:
Will do. But to be clear, the intention of this PR is to stop documenting it as we don't future plugins to use it. We may still want to support it in a more limited capacity for backwards compatibility.

jni:
Thank you! Yes I understand we are removing it but it's good to know like 12 months from now when someone complains to have easy-to-find references to *why* it was removed.

PR #313, opened 2023-09-15
update url for fetching plugin listing from napari hub

nclack:
This PR updates for a couple api changes in dependencies that are blocking #311 and #312.

For the napar-hub api, the `/plugins` endpoint was removed (see [here](https://github.com/chanzuckerberg/napari-hub/discussions/1065)). This impacts the `get_hub_plugins()` function, and a test `test_get_hub_plugins` which fails without this change. It doesn't appear that anything actually uses these, so this PR removes them.

Also [build](https://github.com/pypa/build) updated to 1.0. `IsolatedEnvBuilder` was renamed to `DefaultIsolatedEnv`.

### DIFF
--- a/src/npe2/_inspection/_full_install.py
+++ b/src/npe2/_inspection/_full_install.py
@@ -58,9 +58,9 @@ def isolated_plugin_env(
     """
     # it's important that this import be lazy, otherwise we'll get a circular
     # import when serving as a setuptools plugin with `python -m build`
-    from build.env import IsolatedEnvBuilder
+    from build.env import DefaultIsolatedEnv
 
-    with IsolatedEnvBuilder() as env:
+    with DefaultIsolatedEnv() as env:
         # install the package
         pkg = f"{package}=={version}" if version else package
         logger.debug(f"installing {pkg} into virtual env")
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ classifiers = [
 dependencies = [
     "PyYAML",
     "appdirs",
-    "build",
+    "build>=1",
     "psygnal>=0.3.0",
     "pydantic<2",
     "tomli-w",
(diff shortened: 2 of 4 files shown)


Czaki:
Is it deprecated or removed?

nclack:
> Is it deprecated or removed?

removed. I'll update the description.

Czaki:
Is it impossible to restore it on few months? To have migration time?

nclack:
> Is it impossible to restore it on few months? To have migration time?

I created chanzuckerberg/napari-hub#1263 to request this.

As far as I can tell [1], no one uses the `get_hub_plugins()` function and it's behind a private interface. So I'm not too worried about backward compatibility here. But there are lots of good reasons to use a deprecation window just in case.

Since it looks like nothing uses `get_hub_plugins()` an alternative to this pr is to remove `get_hub_plugins()` and the corresponding test.

[1]: I did a code search through npe2, napari and npe2api.

Czaki:
If it is not used, then it should be removed. The napari-plugin-manager should also be checked.

nclack:
@Czaki I didn't see any use in napari-plugin-manager either. I removed the call and test like you suggested. Also updated the PR description. Let me know if you think of anything else. Getting this in will help #311 and #312 to move forward.

nclack:
@aganders3 can you double-check that we're not using the `get_hub_plugins()` function anywhere?

aganders3:
I also don't see this used anywhere in napari or the plugin manager.

There _was_ a function in the plugin manager using this hub endpoint, but it was not using the npe2 function to reach it anyway. That whole file was removed in napari 0.4.18 and prior to the plugin manager being extracted to its own repo.

src/npe2/_inspection/_full_install.py
```diff
@@ -58,9 +58,9 @@ def isolated_plugin_env(
     """
     # it's important that this import be lazy, otherwise we'll get a circular
     # import when serving as a setuptools plugin with `python -m build`
-    from build.env import IsolatedEnvBuilder
+    from build.env import DefaultIsolatedEnv
```
Czaki, line 61: Could you set the minimum build version in `pyproject.toml`?

PR #319, opened 2023-10-11
Fix tests by disble broken one

Czaki:
In this PR I have disabled the test depending on `example-plugin` python package as it was hidden by the creator.

### DIFF
--- a/tests/test_fetch.py
+++ b/tests/test_fetch.py
@@ -24,6 +24,7 @@ def test_fetch_npe2_manifest():
     assert mf.npe1_shim is False
 
 
+@pytest.mark.skip("package looks deleted from pypi")
 def test_fetch_npe1_manifest_with_writer():
     mf = fetch_manifest("example-plugin")
     assert mf.name == "example-plugin"
--- a/tests/test_fetch.py
+++ b/tests/test_fetch.py
@@ -109,7 +109,7 @@ def test_get_pypi_plugins():
     [
         "https://files.pythonhosted.org/packages/fb/01/e59bc1d6ac96f84ce9d7a46cc5422250e047958ead6c5693ed386cf94003/napari_dv-0.3.0.tar.gz",  # noqa
         "https://files.pythonhosted.org/packages/5d/ae/17779e12ce60d8329306963e1a8dec608465caee582440011ff0c1310715/example_plugin-0.0.7-py3-none-any.whl",  # noqa
-        "git+https://github.com/DragaDoncila/example-plugin.git",
+        # "git+https://github.com/DragaDoncila/example-plugin.git", Draga hide package
         # this one doesn't use setuptools_scm, can check direct zip without clone
         "https://github.com/jo-mueller/napari-stl-exporter/archive/refs/heads/main.zip",
     ],
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,7 @@ jobs:
           python -m pip install -e ./napari-from-github -c "./napari-from-github/resources/constraints/constraints_py3.10.txt"
           python -m pip install -e .[json]
           # bare minimum required to test napari/plugins
-          python -m pip install pytest scikit-image[data] zarr xarray hypothesis matplotlib
+          python -m pip install pytest pytest-pretty scikit-image[data] zarr xarray hypothesis matplotlib
 
       - name: Run napari plugin headless tests
         run: pytest -W 'ignore::DeprecationWarning' napari/plugins napari/settings napari/layers napari/components
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,7 @@ testing = [
     "pytest",
     "pytest-cov",
     "jsonschema",
+    "pytest-pretty",
 ]
 dev = ["black", "ipython", "isort", "mypy", "pre-commit"]
 docs = ["Jinja2", "magicgui>=0.3.3"]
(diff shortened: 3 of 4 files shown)


tests/test_fetch.py
```diff
@@ -24,6 +24,7 @@ def test_fetch_npe2_manifest():
     assert mf.npe1_shim is False
 
 
+@pytest.mark.skip("package looks deleted from pypi")
 def test_fetch_npe1_manifest_with_writer():
     mf = fetch_manifest("example-plugin")
```
Czaki, line 29: @tlambert03 You have created these tests. Did you know who is the author of the `example-plugin` and why it was removed from pypi?

DragaDoncila: Ugh sorry @Czaki! Yeah I removed that plugin because I was playing with other things - sorry, didn't realize tests were using it. I can pop it back up,but if we think this test (and others like it) might be critical, we can think about maintaining a dummy plugin within the napari org.

Czaki: This is the package used for the npe1 test. So I do not think that it will require maintenance. 

Byt we could migrate it if you wish.

brisvag: I think we should definitely migrate it to npe2 and keep it up to date in the napari org!

Czaki: But this is test to check if npe1 shims works.

brisvag: We can have two branches then :P

Czaki: After check we could just create it as npe1, upgrade to npe2 and in tests just use older version

DragaDoncila: Happy to keep such a plugin in the napari org, but just want to make sure we document it and what it's for! @Czaki do you want me to put up the plugin and release its two versions?

Czaki: You mean npe1 and npe2 version? Yes.

Czaki: But of course, we could think about a better name if such a need exists.

DragaDoncila: Yeah agreed, I think it should be `test-dummy-plugin` or something. And we should hide it from hub. Or even not add the framework classifier?

Czaki: I think that not adding a classifier could be a good idea.

DragaDoncila (approved):
@Czaki I'm happy to approve to get at least some green in CI. Once I've got the dummy plugin up (over the weekend most likely), we can come back and update the skipped test.

Czaki:
I do not have permission to merge it. Should I change codecov settings to increase the allowed drop?

tests/test_fetch.py
brisvag: I think we have different ideas here; I was imagining an actual `example-plugin` called such, because I can imagine people looking up "napari example plugin" and it would be an easy way in that's not as "scary" as the cookiecutter.

Maybe this just serves a completely different purpose from the testing, in that case maybe this is not the right conversation :P

nclack:
> Should I change codecov settings to increase the allowed drop?

In the past, for cases like this, we've just forced the merge. I like keeping the codecov settings like they are so that drops are visible.

However, I don't have the permissions to allow the merge. @tlambert03 @DragaDoncila Do you know who does? I'd love @Czaki to get permissions here.

tlambert03:
I don't understand, all napari core devs are on the team here with write access... as are you. 

Is it an admin access thing?

Czaki:
> However, I don't have the permissions to allow the merge. @tlambert03 @DragaDoncila Do you know who does? I'd love @Czaki to get permissions here.

Just check. @tlambert03 is the only one who has such permissions, but any person with admin access to napari organization could add the next such persons. 


> Is it an admin access thing?

yes

tlambert03:
@Czaki Just gave you admin

tests/test_fetch.py
```diff
@@ -108,7 +109,7 @@ def test_get_pypi_plugins():
     [
         "https://files.pythonhosted.org/packages/fb/01/e59bc1d6ac96f84ce9d7a46cc5422250e047958ead6c5693ed386cf94003/napari_dv-0.3.0.tar.gz",  # noqa
         "https://files.pythonhosted.org/packages/5d/ae/17779e12ce60d8329306963e1a8dec608465caee582440011ff0c1310715/example_plugin-0.0.7-py3-none-any.whl",  # noqa
-        "git+https://github.com/DragaDoncila/example-plugin.git",
+        # "git+https://github.com/DragaDoncila/example-plugin.git", Draga hide package
```
nclack, line 112: The code coverage is dropping because this is the only example tested that is fetching using git/git+https. We can cover by replacing this with one of the other examples [here](https://app.codecov.io/gh/napari/npe2/pull/319/blob/src/npe2/_inspection/_fetch.py#L231) e.g. git+https://github.com/tlambert03/napari-dv.git

nclack (changes requested):
Looked into the coverage drop. It is related to fetching plugin info over git or git+https. I added a comment with some more details.

I think I'd prefer it if we add an example or decide to remove the functionality.

Czaki:
@nclack The whole idea is to reenable it within a week but to allow current PR work.

tests/test_fetch.py
DragaDoncila: ooops sorry @brisvag I missed this comment. I think maybe these are two separate things? I always viewed the `cookiecutter` as the default plugin template, but I could imagine maintaining some other example plugin - I'm not sure what you think the benefits of it would be, over the cookiecutter though? 

In general, I think that should be separate to a plugin that we use for testing purposes though since they're different targets. But yeah, happy to consider a richer demo-like plugin!

PR #326, opened 2023-11-21
Fix rendering of manifest reference table.

### LINKED ISSUE #308: Documentation: Improve the fields section of the Manifest reference document
The manifest reference fields section isn't displaying correctly at https://napari.org/dev/plugins/technical_references/manifest.html#fields for two reasons:

1. The part in `icon` should have the three list items below it as part of that table entry. I guess the indentation is off, because the table does not display properly.
2. Above the table it reads; "All fields are optional except those in bold.", but none of the fields are in bold by the looks. I don't know if it is possible to make a bold code block, and if not, maybe this sentence could be removed.

melissawm:
The schema.json file at https://github.com/napari/npe2/releases/latest/download/schema.json also needs to be updated after this PR is merged.

Fixes #308 (not the bold part though, that will only be fixed with the new theme)


DragaDoncila (approved):
Thanks @melissawm! @nclack do you know how to update that schema file?

nclack:
> Thanks @melissawm! @nclack do you know how to update that schema file?

It'll happen automatically. There's a step in the CI.

nclack (approved):
Thanks!

PR #327, opened 2023-11-21
Add constraint for brainreg-segment in test

Czaki:
It looks like brainreg-segment in version 0.2.19 switch to be npe2 plugin


DragaDoncila (approved):
Thanks @Czaki. We should update the dummy plugin and use that for this test instead, but would be good to get CI green again

jni:
LOL the comment a few lines down:

```
    # Test will eventually fail when brainreg-segment is updated to npe2
```

Maybe remove it? 😂

Czaki:
removed

nclack (approved):
👍

PR #336, opened 2024-01-04
replace module separator with path separator

Czaki:
Fix bug reported on zulip https://napari.zulipchat.com/#narrow/stream/309872-plugins/topic/Plugins.20showing.20up.20in.20napari.20plugin.20dialogue/near/411169527

### DIFF
--- a/src/npe2/manifest/_bases.py
+++ b/src/npe2/manifest/_bases.py
@@ -79,7 +79,7 @@ class ImportExportModel(BaseModel):
             try:
                 import tomllib
             except ImportError:
-                import tomli as tomllib
+                import tomli as tomllib # type: ignore [no-redef]
 
             loader = tomllib.load
         elif path.suffix.lower() in (".yaml", ".yml"):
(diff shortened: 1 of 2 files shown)


src/npe2/manifest/_bases.py
```diff
@@ -79,7 +79,7 @@ def from_file(cls, path: Union[Path, str]):
             try:
                 import tomllib
             except ImportError:
-                import tomli as tomllib  # type: ignore [no-redef]
+                import tomli as tomllib
```
Czaki, line 82:
```suggestion
                import tomli as tomllib # type: ignore [no-redef]
```

brisvag (approved):

DragaDoncila (approved):
Thanks @Czaki! I think this is fine to merge without napari tests passing

nclack:
@DragaDoncila @Czaki What do you all think about making a release to get this fix out to pypi?

Czaki:
I think that it is a good idea. The only thing is that I would like to check if new release do not break 0.4.18. It requires half hour for check.

PR #337, opened 2024-01-05
Update fetch mechanism to break only on 404 

Czaki:
I found that when `npe2api` fetch actions happens, not all plugins are collected. This PR add some debug information to `npe2 fetch` used by `npe2api` fetch workflow.


DragaDoncila:
@Czaki would be good to keep coverage at 100% if possible. Happy to merge this otherwise though.

PR #340, opened 2024-03-07
Adopt quotes around type names to make Reader ex. copy/paste-ready

vreuter:
https://github.com/napari/docs/pull/367#pullrequestreview-1921207030
https://github.com/napari/docs/issues/365

### DIFF
--- a/src/npe2/types.py
+++ b/src/npe2/types.py
@@ -60,7 +60,7 @@ Widget = Union["magicgui.widgets.Widget", "qtpy.QtWidgets.QWidget"]
 WidgetCreator = Callable[..., Widget]
 
 # ReaderContribution.command must point to a ReaderGetter
-ReaderFunction = Callable[[PathOrPaths], List["LayerData"]]
+ReaderFunction = Callable[[PathOrPaths], List[LayerData]]
 ReaderGetter = Callable[[PathOrPaths], Optional[ReaderFunction]]
 
 
--- a/src/npe2/types.py
+++ b/src/npe2/types.py
@@ -60,7 +60,7 @@ Widget = Union["magicgui.widgets.Widget", "qtpy.QtWidgets.QWidget"]
 WidgetCreator = Callable[..., Widget]
 
 # ReaderContribution.command must point to a ReaderGetter
-ReaderFunction = Callable[[PathOrPaths], List[LayerData]]
+ReaderFunction = Callable[[PathOrPaths], List["LayerData"]]
 ReaderGetter = Callable[[PathOrPaths], Optional[ReaderFunction]]
 
 
--- a/_docs/example_plugin/some_module.py
+++ b/_docs/example_plugin/some_module.py
@@ -5,11 +5,10 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional
 from magicgui import magic_factory
 from qtpy.QtWidgets import QWidget
 
-from npe2.types import LayerData, PathOrPaths, ReaderFunction
-
 if TYPE_CHECKING:
     import napari.types
     import napari.viewer
+    from npe2.types import LayerData, PathOrPaths, ReaderFunction
 
 
 def write_points(path: str, layer_data: Any, attributes: Dict[str, Any]) -> List[str]:
@@ -20,7 +19,7 @@ def write_points(path: str, layer_data: Any, attributes: Dict[str, Any]) -> List
     return [path]
 
 
-def get_reader(path: PathOrPaths) -> Optional["ReaderFunction"]:
+def get_reader(path: "PathOrPaths") -> Optional["ReaderFunction"]:
     # If we recognize the format, we return the actual reader function
     if isinstance(path, str) and path.endswith(".xyz"):
         return xyz_file_reader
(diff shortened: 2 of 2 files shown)


lucyleeow:
That code snippet certainly gets there magically in a way I don't understand. I did not realise the code was 'copied' over from `_docs/example_plugin/some_module.py`. I am not sure of the purpose/all uses of the `some_module.py` example plugin so now I don't know if we still want to 'string' the annotations there? If we don't want to do strings, it seems we could add import code just above the '# example_plugin.some_module ' comment:

https://github.com/napari/npe2/blob/595b088c81019710b09f1a663b4a09ae34dc1f2a/_docs/templates/_npe2_readers_guide.md.jinja#L33-L36

I note also that the types alias definitions used in annotation of the functions also magically get into the code snippet - BUT we are missing `LayerData`, which is used in annotation but not given.

Maybe @melissawm or @psobolewskiPhD  knows more about this?

src/npe2/types.py
```diff
@@ -60,7 +60,7 @@ def __array__(self) -> "np.ndarray":
 WidgetCreator = Callable[..., Widget]
 
 # ReaderContribution.command must point to a ReaderGetter
-ReaderFunction = Callable[[PathOrPaths], List[LayerData]]
+ReaderFunction = Callable[[PathOrPaths], List["LayerData"]]
```
lucyleeow, line 63: I don't think we want to change the type aliases here.

lucyleeow:
CI failures unrelated, don't worry about them, maybe due to the `importlib` release today?

lucyleeow:
Fix for CIs in https://github.com/napari/npe2/pull/341, we can update with main once that PR is merged. Thanks!

src/npe2/types.py
vreuter: https://github.com/napari/npe2/pull/340/commits/c7236b24597b9be0b7c4758a2acf5f5dff797dca

tlambert03:
I understand the motivation, since [the docs](https://napari.org/stable/plugins/guides.html#reader-example) pluck out sub bits of `example_plugin/some_module.py` that aren't necessarily intended to be copy/pasted without the imports at the top.

But, if you go this direction, then you should also probably update `example_plugin/some_module.py` to put all the type imports inside of `if TYPE_CHECKING`.  As is, this leaves `some_module.py` in a funny looking state with quotes around type names even though they've been imported at the top of the module

```
from npe2.types import LayerData, PathOrPaths, ReaderFunction
```

so, if someone were _only_ looking at that module, they might say "hmm, why did they put them in quotes since they already imported it at the top of the module?"

vreuter:
> I understand the motivation, since [the docs](https://napari.org/stable/plugins/guides.html#reader-example) pluck out sub bits of `example_plugin/some_module.py` that aren't necessarily intended to be copy/pasted without the imports at the top.
> 
> But, if you go this direction, then you should also probably update `example_plugin/some_module.py` to put all the type imports inside of `if TYPE_CHECKING`. As is, this leaves `some_module.py` in a funny looking state with quotes around type names even though they've been imported at the top of the module
> 
> ```
> from npe2.types import LayerData, PathOrPaths, ReaderFunction
> ```
> 
> so, if someone were _only_ looking at that module, they might say "hmm, why did they put them in quotes since they already imported it at the top of the module?"

indeed, good point @tlambert03 -- i update here: 3de817819c283bde22fccd4ad2e4c82d3bae3669

vreuter:
@tlambert03 I guess the CI depends on #344 for the moment?

tlambert03:
Looks like it

lucyleeow:
@tlambert03 any idea how the type aliases at the top of the code snippet get there?

image

It does not seem to be from the code block in the jinja template:

https://github.com/napari/npe2/blob/595b088c81019710b09f1a663b4a09ae34dc1f2a/_docs/templates/_npe2_readers_guide.md.jinja#L33-L36

lucyleeow (approved):
Thanks! LGTM!

tlambert03:
> @tlambert03 any idea how the type aliases at the top of the code snippet get there?
> It does not seem to be from the code block in the jinja template:
> ```python 
># example_plugin.some_module 
>{{ 'readers'|example_implementation }} 
>``` 

it is actually from that jinja block.  that pipe (`|`) character in the `{{ 'readers'|example_implementation }}` expression is a [jinja filter](https://jinja.palletsprojects.com/en/3.1.x/templates/#filters).  It's a function defined somewhere that can modify the arguments passed to it (the bit before the pipe, in this case `readers`).   If you grep for `example_implementation` in the repo, you'll find the function here:

https://github.com/napari/npe2/blob/595b088c81019710b09f1a663b4a09ae34dc1f2a/_docs/render.py#L125-L130


which calls:

https://github.com/napari/npe2/blob/595b088c81019710b09f1a663b4a09ae34dc1f2a/_docs/render.py#L94

which grabs the needed types:

https://github.com/napari/npe2/blob/595b088c81019710b09f1a663b4a09ae34dc1f2a/_docs/render.py#L117C14-L122

which is defined here:

https://github.com/napari/npe2/blob/595b088c81019710b09f1a663b4a09ae34dc1f2a/_docs/render.py#L79-L91

DragaDoncila (approved):
CI is green again after merging main. Thanks for these updates @vreuter ❤️

PR #341, opened 2024-03-08
TST Fix `test_conversion_missing`

lucyleeow:
This test is failing in CIs in #340.

I think it is because there is no warning raised (not sure why the `UserWarning` is there). It used to pass because of this bug in pytest https://github.com/pytest-dev/pytest/issues/9036, but this was fixed in https://github.com/pytest-dev/pytest/pull/11129 and included in a recent release, thus our CIs are failing. 

Note that I have update the test to be more specific about the exception raised (`PackageNotFoundError`) but it is not needed, subclass exceptions are still accepted by pytest. Also added a message to be more specific.

### DIFF
--- a/tests/test_conversion.py
+++ b/tests/test_conversion.py
@@ -141,14 +141,16 @@ setup(
 
 def test_conversion_missing():
     with pytest.raises(
-        PackageNotFoundError, match="No package or entry point found with name",
+        PackageNotFoundError,
+        match="No package or entry point found with name",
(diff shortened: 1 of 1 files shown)


lucyleeow:
cc @DragaDoncila

DragaDoncila (approved):
Yep, looks great to me thanks @lucyleeow

jni (approved):

PR #345, opened 2024-04-03
DOC Add note on string type annotations in widget contributions

lucyleeow:
Add note that type annotations should be str to avoid having `napari` as dependency


lucyleeow:
cc @DragaDoncila

DragaDoncila (approved):
Looks good!

psobolewskiPhD (approved):
Thanks!

PR #348, opened 2024-05-28
Add dictionary mapping `command_id` to its associated `MenuCommand`s for each plugin

DragaDoncila:
Adding this dictionary to support quick mapping of widgets/commands to their menus for implementing contributable menus . Prior to this PR, you would have to search all menu contributions for the command of the contribution you were currently processing. This map allows you direct access to the menus this command needs to live in. 

This is a precursor to potentially rearchitecting the manifest schema entirely to be "command-first".

### DIFF
--- a/src/npe2/_plugin_manager.py
+++ b/src/npe2/_plugin_manager.py
@@ -623,7 +623,7 @@ class PluginManager:
         for mf in self.iter_manifests(disabled=disabled):
             yield from mf.contributions.menus.get(menu_key, ())
 
-    def menus(self, disabled=False) -> Dict[str, List[MenuCommand]]:
+    def menus(self, disabled=False) -> Dict[str, List[MenuItem]]:
         """Return all registered menu_key -> List[MenuItems]."""
         _menus: DefaultDict[str, List[MenuItem]] = DefaultDict(list)
         for mf in self.iter_manifests(disabled=disabled):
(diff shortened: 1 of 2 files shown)


src/npe2/_plugin_manager.py
```diff
@@ -358,14 +362,36 @@ def register(
...
+    def _get_associated_menus(
+        self, manifest: PluginManifest, command_id: str
+    ) -> Dict[str, List[MenuCommand]]:
+        menus = manifest.contributions.menus or dict()
+        associated_menus = defaultdict(list)
+        for menu_id, items in menus.items():
+            for item in items:
+                if getattr(item, "command", "") == command_id:
+                    associated_menus[menu_id].append(item)
+        return associated_menus
```
jni, line 386: Ah this is backwards. You should be able to do this in a single pass of the menus:
- iterate over the menus, but
- map commands back to menus
- since you use a defaultdict it's ok if some menu-less commands are missed, but you could add them by keeping a set of already-seen commands and then putting in the missed ones at the end.
 
So:

```suggestion
        menu_map = self._command_menu_map[manifest.name]  # just for conciseness below
        for menu_id, menu_items in manifest.contributions.menus or ():
            # command IDs are keys in map
            # each value is a dict menu_id: list of MenuCommands
            # for the command and menu
            for item in menu_items:
                if command_id := getattr(item, "command", None) is not None:
                    menu_map[command_id].append(item)
```

jni (approved):

PR #355, opened 2024-06-21
Add registration to plugin categories

### LINKED ISSUE #317: Proposal: add "registration" to allowed categories
I just read through #239 (sorry for repeatedly ignoring pings there 😂). I understand that registration is "intended" to go under transformations but frankly I found it confusing. I'd pull it out into its own category. What is a registration vs a transformation? Something that involves two or more input images, rather than a single image.

jni:
Closes #317

### DIFF
--- a/src/npe2/manifest/schema.py
+++ b/src/npe2/manifest/schema.py
@@ -29,7 +29,7 @@ from .utils import Executable, Version
 logger = getLogger(__name__)
 
 
-SCHEMA_VERSION = "0.2.0"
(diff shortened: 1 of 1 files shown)


jni:
@DragaDoncila? 🙏

DragaDoncila (approved):
Yep, I think Registration is definitely fair to include, especially as we now have a menu for it yay! I don't think this is sufficiently controversial to warrant a waiting period, so I'm going to merge.

PR #356, opened 2024-07-28
On disable, dont try to pop a key that doesn't exist

psobolewskiPhD:
Closes https://github.com/napari/napari/issues/7135

This is a naive fix. If the key doesn't exist when disabling, just return None.

### DIFF
--- a/src/npe2/_plugin_manager.py
+++ b/src/npe2/_plugin_manager.py
@@ -486,7 +486,7 @@ class PluginManager:
 
         self._disabled_plugins.add(plugin_name)
         self._contrib.remove_contributions(plugin_name)
-        self._command_menu_map.pop(plugin_name)
+        self._command_menu_map.pop(plugin_name, None)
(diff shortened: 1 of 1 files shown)


DragaDoncila:
This might fix the issue but I'm concerned that the early `return` in [that function](https://github.com/napari/npe2/pull/356/files#diff-ba77afb51bb2016d47d6b165aa031ca15029f3f38615bfa32df7808e03db78ffL481) didn't save us. I was also able to reproduce this back to 0.4.18 which seems unbelievable?? So I'm looking at this a bit closer to figure out what's going on

DragaDoncila:
Ahh yess, I see what's going on. It's my own code 🤣 And that's why it's broken back to 0.4.18, because it's in the new npe2 release...

DragaDoncila (approved):
Confirmed this addresses the issue. Let's merge and release npe2 asap

PR #364, opened 2024-10-24
Change npe1 plugin used for sample data test

### LINKED ISSUE #363: CI failing due to test expecting package name with hyphens containing underscores instead
### Description

See #362 for an example failure, but TLDR:

Our `test_fetch_npe1_manifest_with_sample_data` [test](https://github.com/napari/npe2/blob/main/tests/test_fetch.py#L37) expects `napari-pyclesperanto-assistant` to have hyphens in the name, but as of releasing v0.23.0 on October 18th, the package's metadata has underscores in the name.

The [package name](https://pypi.org/project/napari-pyclesperanto-assistant/#history) on `pypi` still has hyphens, even though [the package's setup.py](https://github.com/clEsperanto/napari_pyclesperanto_assistant/blob/master/setup.py#L33) file has underscores. I think that's because of [PEP503](https://packaging.python.org/en/latest/specifications/name-normalization/), but either `setuptools_scm` or some other package in the middle is no longer normalizing names I guess?

None of this is super relevant to us I guess, except that I'm curious what's changed. Based on [this comment](https://github.com/napari/npe2/blob/main/src/npe2/manifest/schema.py#L95) in the schema, we are not normalizing package names for the manifest name, so the underscore is fine in that regard. It's just a little strange because it technically would no longer match the package name as it's listed on pypi - note though that `fetch_manifest` works if it's passed underscores **or** hyphens.

Is it fine to just change the test to use underscores? That's what I'm leaning towards here, but I'm not sure if there will be gotchas about this somewhere down the line... @jni @Czaki? We could also just use any of the other sample-data providing manifests and avoid this entirely!

DragaDoncila:
Closes #363 by changing the plugin used in the test to avoid name normalization issues entirely.

### DIFF
--- a/tests/test_fetch.py
+++ b/tests/test_fetch.py
@@ -36,7 +36,7 @@ def test_fetch_npe1_manifest_with_sample_data():
     mf = fetch_manifest("napari-kics")
     assert mf.name == "napari-kics"
     assert mf.contributions.sample_data
-    # Test will eventually fail when napari-pyclesperanto-assistant is updated to npe2
+    # Test will eventually fail when napari-kics is updated to npe2
     # This is here as a sentinel
     assert mf.npe1_shim is True
 
--- a/tests/test_fetch.py
+++ b/tests/test_fetch.py
@@ -33,8 +33,8 @@ def test_fetch_npe1_manifest_with_writer():
 
 
 def test_fetch_npe1_manifest_with_sample_data():
-    mf = fetch_manifest("napari-pyclesperanto-assistant")
-    assert mf.name == "napari-pyclesperanto-assistant"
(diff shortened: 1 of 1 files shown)


jni:
@DragaDoncila I think you should still normalise (use `.replace('_', '-')`) because nothing at all is stopping that package from updating in the same broken way.

DragaDoncila:
@jni if we want to normalise, I'm happy to, but I don't think randomly changing this single test is the way to do it. Or even changing every test for that matter.

I think we should get this in to fix CI, and then have a separate discussion about whether we want to normalise manifest names, and when.

Czaki (approved):

PR #368, opened 2025-01-01
Tolerate distributions with no Name

cjwatson:
https://bugs.debian.org/1090247 shows a napari test failure caused by this on Python 3.13.  On closer inspection I found that `/usr/lib/python3.13/dist-packages/coverage-7.6.0.egg-info` and `/usr/lib/python3.13/dist-packages/tqdm-4.67.1.dist-info` existed with no `Name` metadata field.  This shouldn't cause a failure here.

### DIFF
--- a/src/npe2/manifest/utils.py
+++ b/src/npe2/manifest/utils.py
@@ -81,7 +81,13 @@ class Executable(GenericModel, Generic[R]):
 
             # look for a package name in the command id
             dists = sorted(
-                (d.metadata["Name"] for d in distributions()), key=len, reverse=True
+                (
+                    d.metadata["Name"]
+                    for d in distributions()
+                    if d.metadata["Name"] is not None
+                ),
+                key=len,
+                reverse=True,
             )
             for name in dists:
                 if self.command.startswith(name):  # pragma: no cover


jni:
Thanks for the PR @cjwatson! You've caught us at an awkward time because npe2api relied on web scraping PyPI search results for some functionality, and that was [recently broken by PyPI due to abuse](https://discuss.python.org/t/fastly-interfering-with-pypi-search/73597). I think this is why the CI is failing. We're still figuring out the best fix but it'll probably be minimum 1-2 weeks until a fix because of the Christmas/New Year's break. 🙏

jni:
For other devs stumbling on this, this is the relevant bit of the report:

```
self = ReaderContribution(command='tmp_plugin._', filename_patterns=['*.fake'], accepts_directories=False)

    @property
    def plugin_name(self) -> str:
        """Return the manifest/plugin name for this contribution."""
        if not self._plugin_name:
            # we will likely never get here if the contribution was created
            # as a member of a PluginManifest.
            # But if not, we can use a couple heuristics...
            from importlib.metadata import distributions
    
            # look for a package name in the command id
>           dists = sorted(
                (d.metadata["Name"] for d in distributions()), key=len, reverse=True
            )
E           TypeError: object of type 'NoneType' has no len()
```

DragaDoncila:
OK the reason we would ever be checking the `plugin_name` of `tqdm` here is because of the `discover` [utility](https://github.com/napari/npe2/blob/f99a77a98b78c0d2a7c54f60a77ef96963b253bf/npe2/manifest/schema.py#L255) which goes through installed packages looking for npe2 manifests (or npe1 entrypoints). The effect of this change, I believe, is that if a napari plugin does not declare a `name` in its distribution metadata, it would not be discovered as a plugin. I think this is ok, and likely to be an extreme edge case, so I'm happy to merge the change as-is.

jni (approved):

PR #369, opened 2025-01-15
Instantiate `_command_menu_map` with all needed `defaultdicts`

DragaDoncila:
This PR fixes a bug in `napari/napari` with loading shimmed npe1 plugins.

On the napari side, we never distinguish between shimmed npe1 plugins and npe2 plugins. This means even though npe1 plugins can't declare menu contributions, we need to be able to index into the `_command_menu_map` for any commands shimmed plugins declare, and get an empty list back.

This PR moves the instantiation of the two extra levels of `defaultdict` up to the declaration of the `_command_menu_map`, so that it can always be fully indexed. I've also added an additional comment, and a specific test for npe1 adapters.

### DIFF
--- a/src/npe2/_plugin_manager.py
+++ b/src/npe2/_plugin_manager.py
@@ -237,7 +237,9 @@ class PluginManager:
         self.events = PluginManagerEvents(self)
         self._npe1_adapters: List[NPE1Adapter] = []
         self._command_menu_map: Dict[str, Dict[str, Dict[str, List[MenuCommand]]]] = (
-            defaultdict(dict)
+            # for each manifest, maps command IDs to menu IDs to list of MenuCommands
+            # belonging to each menu
+            defaultdict(lambda: defaultdict(lambda: defaultdict(list)))
         )
 
         # up to napari 0.4.15, discovery happened in the init here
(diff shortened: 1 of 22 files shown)


src/npe2/_plugin_manager.py
```diff
@@ -237,7 +237,15 @@ def __init__(
...
         self._command_menu_map: Dict[str, Dict[str, Dict[str, List[MenuCommand]]]] = (
-            defaultdict(dict)
+            # for each manifest, maps command IDs to menu IDs to list of MenuCommands
+            # belonging to that menu
+            # i.e. manifest -> command -> menu_id -> list[MenuCommand]
+            # we keep a list of MenuCommands even though we've already indexed by
+            # command ID **and** menu ID because it's possible to declare the
+            # same command in the same menu with different `when` and `group`
+            # fields. This seems an unlikely use case, but it is possible in
+            # the current schema.
```
jni, line 247: Love the comment! I started reading it and found the mapping (line 242) and was like ooh boy I've forgotten this already why didn't I review when she pushed 😂, and by the time I finished it I was like, boom, that's right, this is great! 😃

jni (approved):

PR #374, opened 2025-02-19
Fix naming of npe1 converted contribution commands and add tests

DragaDoncila:
Prior to this PR, declaring multiple implementations of the same hookspec in an npe1 plugin would lead to a converted manifest with multiple commands of the same ID e.g. `plugin_name.get_reader` multiple times, or `plugin_name.write_labels` multiple times.

Even though this is technically "valid" per the schema, it should not be, because this leads to errors in trying to register the commands in napari. App-model (correctly, imo), does not allow you to register multiple commands with the same ID.

We should update the schema validator to disallow multiple declarations of the same command ID, but in the meantime, this PR disambiguates converted commands for the same hook spec by using the function name instead of the spec name.

### DIFF
--- a/src/npe2/_inspection/_from_npe1.py
+++ b/src/npe2/_inspection/_from_npe1.py
@@ -404,11 +404,11 @@ class HookImplParser:
         )
 
     def add_command(self, impl: HookImplementation, py_name: str = "") -> str:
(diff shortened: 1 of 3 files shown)


jni (approved):
wow, nice find 😵

PR #379, opened 2025-05-13
Clarify that sample data contributions must return valid layer data

### LINKED ISSUE #378: `fetch_manifest` tests failing when building package from github zip
In #377 our CI is [failing](https://github.com/napari/npe2/actions/runs/14941620192/job/41979543684?pr=377#step:6:93) for `python >= 3.9` on the `fetch_manifest` [test](https://github.com/napari/npe2/blob/main/tests/test_fetch.py#L95) that downloads a `zip` directly from github and then builds the package to extract the manifest.

The [plugin in question](https://github.com/jo-mueller/napari-stl-exporter) has not changed since 2023. I've narrowed this down to a `setuptools` issue: package includes the manifest with `setuptools==80.0.1` but doesn't with `setuptools==80.1.0`. Looking at the [changelog](https://github.com/pypa/setuptools/compare/v80.0.1...v80.1.0) for `setuptools` between these two versions, it seems likely this is related to the `setup.py` deprecation and possibly [this PR](https://github.com/pypa/setuptools/pull/4970)? But I don't know enough about these things to fully grok what's happening.

The package builds without error locally, and includes all the necessary code for the plugin, it just doesn't include the manifest.

According to the comment on the test, we use this specific plugin because it doesn't rely on `setuptools.scm` for versioning, and that means we can download the zip directly without cloning. 

Without knowing exactly what's causing the issue, I guess our options are:

- Remove this test and the zip download functionality: the standard way we fetch manifests in `npe2api` is by grabbing the wheel directly from PyPI, so I don't think there's a danger of not testing this functionality breaking `npe2api`. I think this was only implemented at all as a "completeness" measure, rather than because it's commonly used.
- Pin the `setuptools` version for this test: I don't actually know exactly how to do this but I assume there's a way to declare a specific setuptools version for this test only.
- Update the plugin (somehow): presumably we could fix this somehow by moving away from `setup.py` in the plugin while still maintaining the dependency on `setuptools` and getting the test to pass with the latest `setuptools` version

I'm tempted to just remove this test and functionality, but if there's an easy fix out there, then, sure. @Czaki you're probably the man with the answers here.

DragaDoncila:
Follow on from [napari/#7851](https://github.com/napari/napari/pull/7851) I thought we should clarify that sample contributions are expected to always return valid layer data, unlike readers which can return the null layer sentinel. 

Note we'll need to resolve #378 for CI to pass.

### DIFF
--- a/_docs/templates/_npe2_sample_data_guide.md.jinja
+++ b/_docs/templates/_npe2_sample_data_guide.md.jinja
@@ -10,6 +10,10 @@ It can take the form of a **Sample Data URI** that points to a static resource
 (such as a file included in the plugin distribution, or a remote resource),
 or **Sample Data Function** that generates layer data on demand.
 
+Note that unlike reader contributions, sample data contributions are
+**always** expected to return data, so returning `[(None,)]`
+will cause an error.
+
 ### Sample Data example
 
 ::::{tab-set}
--- a/src/npe2/manifest/contributions/_sample_data.py
+++ b/src/npe2/manifest/contributions/_sample_data.py
@@ -32,7 +32,9 @@ class SampleDataGenerator(_SampleDataContribution, Executable[List[LayerData]]):
     """Contribute a callable command that creates data on demand."""
 
     command: str = Field(
-        ..., description="Identifier of a command that returns layer data tuple."
+        ...,
+        description="Identifier of a command that returns layer data tuple. "
+        "Note that this command cannot return `[(None,)]`.",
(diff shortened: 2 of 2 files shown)


brisvag (approved):

psobolewskiPhD (approved):

psobolewskiPhD:
Note this is mostly to get documentation to be correct.

willingc (approved):

PR #382, opened 2025-06-25
Replace appdirs with platformdirs and silence importlib warning in tests

klmurussi:
This PR includes three small and isolated commits extracted from Debian packaging efforts for `npe2`. These changes were originally introduced as patches in the `debian/patches` directory, but they are general improvements and should be
relevant upstream as well.

### Summary of changes:

1. **Replace appdirs with platformdirs in imports**
   The `appdirs` module is deprecated in favor of `platformdirs`, which is more actively maintained and has better cross-platform support.

2. **Update `pyproject.toml` to use `platformdirs`**
   This reflects the change above in the declared dependencies.

3. **Silence importlib's DeprecationWarning in test**
   Python emits a deprecation warning for implicit `None` return values.
   This patch silences that specific warning in `test_discover_errors` to avoid noise and test pollution. This was particularly important in Debian's CI where test suites may fail on unhandled warnings.

### Why upstream?

These were applied as part of the Debian packaging process, but they reflect
general improvements:
- Aligning with modern dependencies (`platformdirs`)
- Keeping test output clean
- Removing need for downstream-only patches

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
           python -m pip install pytest pytest-pretty scikit-image[data] zarr xarray hypothesis matplotlib
 
       - name: Run napari plugin headless tests
-        run: pytest -W 'ignore::DeprecationWarning' napari/plugins napari/settings napari/layers napari/components
+        run: pytest -W 'ignore::DeprecationWarning' src/napari/plugins src/napari/settings src/napari/layers src/napari/components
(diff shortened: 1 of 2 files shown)


psobolewskiPhD:
Thanks for the PR! Our CI is having some issues here, not sure what's up.
@DragaDoncila any ideas?

DragaDoncila (approved):
Looks great, thanks @klmurussi :pray:

PR #384, opened 2025-06-26
Fix CI by skipping test and adding `src/` to napari test paths

DragaDoncila:
Since we have some PRs in-flight I'm getting  CI green again by:

- skipping the test causing the failure described in #378 so that we can get CI green. In future we should find a different plugin to test.
- adding `src/` to the napari test paths

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
           python -m pip install pytest pytest-pretty scikit-image[data] zarr xarray hypothesis matplotlib
 
       - name: Run napari plugin headless tests
-        run: pytest -W 'ignore::DeprecationWarning' napari/src/plugins napari/src/settings napari/src/layers napari/src/components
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -83,7 +83,7 @@ jobs:
           python -m pip install pytest pytest-pretty scikit-image[data] zarr xarray hypothesis matplotlib
 
       - name: Run napari plugin headless tests
-        run: pytest -W 'ignore::DeprecationWarning' napari/plugins napari/settings napari/layers napari/components
+        run: pytest -W 'ignore::DeprecationWarning' napari/src/plugins napari/src/settings napari/src/layers napari/src/components
```
psobolewskiPhD, line 86:
```suggestion
        run: pytest -W 'ignore::DeprecationWarning' src/napari/plugins src/napari/settings src/napari/layers src/napari/components
```
I think you meant this way? since the code is under napari in the src dir?

DragaDoncila: Shhh you didn't see anything :dotted_line_face:

psobolewskiPhD (approved):

PR #383, opened 2025-06-26
Add guide for menu contributions

Modjular:
Added a small guide for the new menu contributions feature. Prompted by [this forum thread](https://forum.image.sc/t/npe2-add-plugin-commands-to-the-viewer-menu/113440/6)

Related: https://github.com/napari/docs/issues/585

PR #385, opened 2025-06-26
Update ci.yml to build docs and upload artifact

psobolewskiPhD:
It looks like we have a jupyter book config and toc here, so lets build it and upload the artifact so we can look at a preview.


psobolewskiPhD:
@DragaDoncila 

Ok this is working. it runs the render script and then builds the whole plugins folder .md into html.
So I think this would render the e.g. new file added in #383 ?

docs/_config.yml
```diff
@@ -45,7 +45,8 @@ sphinx:
     html_theme_options: {}
     pygments_style: solarized-dark
     suppress_warnings: ["myst.header"]
-    exclude_patterns: ['**/_*.md']  # includes
+    # build the generated files in this repo to preview them
```
DragaDoncila, line 48: Should we just delete both lines here?

DragaDoncila (approved):
This looks good to me - got confused about the contributions reference cause it was highlighted but I still had to click it. Super nbd. I think let's get this in so we can merge into #383 and see that rendered!

docs/_config.yml
psobolewskiPhD: I commented in case we want to go back to the old way or do something different

PR #386, opened 2025-06-27
Update _contributions.py to show docs for menus and submenus

psobolewskiPhD:
Depends on: https://github.com/napari/npe2/pull/383

Companion PR to https://github.com/napari/npe2/pull/383 this should show the docs in the autogenerated section.
Would be nice to have https://github.com/napari/npe2/pull/385 to confirm.


src/npe2/manifest/contributions/_contributions.py
```diff
@@ -36,8 +36,8 @@ class ContributionPoints(BaseModel):
     widgets: Optional[List[WidgetContribution]]
     sample_data: Optional[List[SampleDataContribution]]
     themes: Optional[List[ThemeContribution]]
-    menus: Dict[str, List[MenuItem]] = Field(default_factory=dict, hide_docs=True)
-    submenus: Optional[List[SubmenuContribution]] = Field(None, hide_docs=True)
+    menus: Dict[str, List[MenuItem]] = Field(default_factory=dict, hide_docs=False)
+    submenus: Optional[List[SubmenuContribution]] = Field(None, hide_docs=False)
```
TimMonko, line 40: I don't fully understand, but wouldn't it be possible to take out the Field assignment? since hide_docs is True by default
```suggestion
    menus: Dict[str, List[MenuItem]]
    submenus: Optional[List[SubmenuContribution]]
```

psobolewskiPhD:
I think it's failing earlier. not sure where the description comes from.

psobolewskiPhD:
I gave fixing this a go with some Copilot help. It's not looking good.
I don't have time to debug more today, but basically the structure of the menus/submenus contribution is sufficiently different that neither render.py nor the template handle it correctly.
Here's the rendered output
image

You can see that menus snippet is fetched correctly, but submenus is not. The null then breaks the toml so I disabled it to debug.
I think this PR will need to be started over with a better understanding of what is being done, rather than my chain patching here.

DragaDoncila:
@melissawm this is the PR we discussed in docs wg where the autogenerated guide is not working

DragaDoncila:
@willingc if you'd like to take a look as well

PR #389, opened 2025-07-04
Add test for get manifest from zip and improve logic to recurse the zip to find manifest

willingc:
This PR tests the function and brings coverage back to green.

  The test uses mocks and verifies that `_get_manifest_from_zip_url`:
  - Properly uses the context manager
  - Correctly finds the first directory inside the extracted path
  - Passes that directory to the build function and returns its result

This PR modifies the `_get_manifest_from_zip_url` to recurse the zip beyond the first directory to look for the manifest. Keeps the original approach as a fallback if searching the zip does not find a manifest.


willingc:
@psobolewskiPhD @DragaDoncila This PR fixes the logic that was causing the zip fetch to fail. Back to 100% coverage. Need to add test to cover a few of the added lines.

PR #414, opened 2025-10-28
Expand reader guide to provide more information about the `get_reader` function

DragaDoncila:
This PR updates the reader guide with more detailed information about how we determine whether a plugin is compatible with a given path, and provides some examples of what the `get_reader` function could do to ensure compatibility.

**PS-** Our CI was failing because of a typer update. Previously, passing `None` to an optional field (e.g. `npe2 list --fields None` replaced the `None` with the default value, but now the `None` is passed through. I therefore update the test to explicitly pass the default value for `fields`.


TimMonko (approved):
🚀 ⛵

PR #434, opened 2026-02-10
chore: add testing against minimum requirements

Czaki:


pyproject.toml
```diff
@@ -32,16 +32,16 @@ classifiers = [
...
 ]
 dependencies = [
-    "PyYAML",
-    "platformdirs",
+    "PyYAML>=6.0",
+    "platformdirs>=4.3.0",
     "build>=1",
     "psygnal>=0.3.0",
     "pydantic>2",
     "pydantic_extra_types",
```
psobolewskiPhD: this also needs a minimum, because the initial was just a stub
```suggestion
    "pydantic_extra_types>2",
```

.github/workflows/ci.yml
```diff
@@ -23,37 +23,45 @@ jobs:
...
           python-version: ${{ matrix.python-version }}
+          activate-environment: true
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
-          pip install -e .[json,docs,testing]
+          uv pip install -e .[testing,docs]
+        env:
+          UV_RESOLUTION: ${{ matrix.resolution }}
```
DragaDoncila: what does this do/why is it needed?

DragaDoncila (approved):
LGTM! Left a comment for my own understanding.

.github/workflows/ci.yml
Czaki: This change is which packages uv install. The `highest` means the newest available that satisfies constraints. The `lowest-direct` should install direct dependencies in the lowest possible version. 

There is also `lowest` that tries to install all packages in the lowest possible version, but it requires that all dependencies correctly point to the lowest supported version.

PR #450, opened 2026-04-03
Unify icon contributions and clarify supported formats

brisvag:
This PR aims to unify a bit the manifest regarding icons, and to bring it more in line with what the underlying app-model accepts.

We actually "officially" support more things than appmodel (e.g: https urls), and I'm not sure whether we plan to keep these and handle the extra logic ourselves or what else to do. At least with this we start supporting dark/light for all icon fields.

### DIFF
--- a/src/npe2/manifest/_validators.py
+++ b/src/npe2/manifest/_validators.py
@@ -1,7 +1,5 @@
 import re
 
-from npe2.manifest.contributions._icon import Icon
-
 _package_name = "([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])"
 _python_identifier = "([a-zA-Z_][a-zA-Z_0-9]*)"
 
@@ -66,24 +64,4 @@ def display_name(v: str) -> str:
             "and must not begin or end with an underscore, white space, or "
             "non-word character."
         )
-    return v
-
-
-def icon_path(v: str | Icon | None) -> str | Icon | None:
-    if isinstance(v, str) and v.startswith("http"):
-        if not v.startswith("https://"):
-            raise ValueError(
-                f"{v} is not a valid icon URL. It must start with 'https://'"
-            )
-    if isinstance(v, Icon):
-        if v.light is not None and v.light.startswith("http"):
-            if not v.light.startswith("https://"):
-                raise ValueError(
-                    f"{v.light} is not a valid icon URL. It must start with 'https://'"
-                )
-        if v.dark is not None and v.dark.startswith("http"):
-            if not v.dark.startswith("https://"):
-                raise ValueError(
-                    f"{v.dark} is not a valid icon URL. It must start with 'https://'"
-                )
     return v
--- a/src/npe2/manifest/_validators.py
+++ b/src/npe2/manifest/_validators.py
@@ -65,3 +65,25 @@ def display_name(v: str) -> str:
             "non-word character."
         )
     return v
+
+
+def coerce_icon(cls, value):
+    if value is None:
+        return None
+    if isinstance(value, str) and value.startswith("http"):
+        if not value.startswith("https://"):
+            raise ValueError(
+                f"{value} is not a valid icon URL. It must start with 'https://'"
+            )
+    # aftervalidator, so it's guaranteed to be of type Icon
+    if value.light is not None and value.light.startswith("http"):
+        if not value.light.startswith("https://"):
+            raise ValueError(
+                f"{value.light} is not a valid icon URL. It must start with 'https://'"
+            )
+    if value.dark is not None and value.dark.startswith("http"):
+        if not value.dark.startswith("https://"):
+            raise ValueError(
+                f"{value.dark} is not a valid icon URL. It must start with 'https://'"
+            )
+    return value
--- a/src/npe2/manifest/_validators.py
+++ b/src/npe2/manifest/_validators.py
@@ -1,5 +1,7 @@
 import re
 
+from npe2.manifest.contributions._icon import Icon
+
 _package_name = "([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])"
 _python_identifier = "([a-zA-Z_][a-zA-Z_0-9]*)"
 
@@ -67,14 +69,21 @@ def display_name(v: str) -> str:
     return v
 
 
-def icon_path(v: str) -> str:
-    if not v:
-        return ""
-    if v.startswith("http"):
+def icon_path(v: str | Icon | None) -> str | Icon | None:
+    if isinstance(v, str) and v.startswith("http"):
         if not v.startswith("https://"):
             raise ValueError(
                 f"{v} is not a valid icon URL. It must start with 'https://'"
             )
-        return v
-    assert isinstance(v, str), f"{v} must be a string"
+    if isinstance(v, Icon):
+        if v.light is not None and v.light.startswith("http"):
+            if not v.light.startswith("https://"):
+                raise ValueError(
+                    f"{v.light} is not a valid icon URL. It must start with 'https://'"
+                )
+        if v.dark is not None and v.dark.startswith("http"):
+            if not v.dark.startswith("https://"):
+                raise ValueError(
+                    f"{v.dark} is not a valid icon URL. It must start with 'https://'"
+                )
     return v
--- a/src/npe2/manifest/_validators.py
+++ b/src/npe2/manifest/_validators.py
@@ -67,24 +67,20 @@ def display_name(v: str) -> str:
     return v
 
 
+def _ensure_valid_https(value):
+    if value.startswith("http") and not value.startswith("https://"):
+        raise ValueError(
+            f"{value} is not a valid icon URL. It must start with 'https://'"
+        )
+
+
 def coerce_icon(value):
     if value is None:
         return None
-    if isinstance(value, str) and value.startswith("http"):
-        if not value.startswith("https://"):
-            raise ValueError(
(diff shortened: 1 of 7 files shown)


src/npe2/manifest/_validators.py
```diff
@@ -1,5 +1,7 @@
 import re
 
+from npe2.manifest.contributions._icon import Icon
```
willingc: @brisvag I think that this line is causing the circular import.

willingc (commented):
Looks good. Seems to have a circular import that in the tests.

brisvag:
Since it's not reused, I just moved the validator to the model where it's needed.

src/npe2/manifest/schema.py
```diff
@@ -122,17 +123,21 @@ class PluginManifest(ImportExportModel):
...
+    icon: str | Icon | None = Field(
+        None,
+        description="Icon used to represent this plugin in the UI, on"
+        " buttons or in menus. Can be a single string or two different options"
+        " for light and dark themes. These values may be:"
+        "<ul><li> a secure (https) URL </li>"
         "<li>a string in the format `{package}:{resource}`, where `package` and "
-        "`resource` are arguments to `importlib.resources.path(package, resource)`, "
-        "(e.g. `top_module.some_folder:my_logo.png`).</li></ul>",
+        "`resource` are arguments to `importlib.resources.files(package, resource)` "
```
psobolewskiPhD, line 133: BTW Is this right? Looking at the docs `files` can only accept one argument, not two:
https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files

src/npe2/manifest/contributions/_commands.py
```diff
@@ -73,11 +73,18 @@ class CommandContribution(BaseModel):
     )
     icon: str | Icon | None = Field(
```
psobolewskiPhD, line 74: should there be validation here, e.g. on the url for https vs http like in the PluginManifest.icon?

src/npe2/manifest/contributions/_submenu.py
```diff
@@ -17,9 +17,16 @@ class SubmenuContribution(BaseModel):
     )
     icon: str | Icon | None = Field(
```
psobolewskiPhD, line 18: should there be validation here, e.g. on the url for https vs http like in the PluginManifest.icon?

src/npe2/manifest/contributions/_commands.py
brisvag: ah, damn, my bad... So it needs to be reused :/

src/npe2/manifest/schema.py
brisvag: oops, forgot to revert this after I realized how it worked.

tests/test_manifest.py
```diff
@@ -182,7 +183,19 @@ def test_visibility():
 
 
 def test_icon():
-    PluginManifest(name="myplugin", icon="my_plugin:myicon.png")
+    pm = PluginManifest(name="myplugin", icon="my_plugin:myicon.png")
+    assert pm.icon == "my_plugin:myicon.png"
+    pm = PluginManifest(name="myplugin", icon="https://example.com/icon.png")
```
Czaki: I don't think that we should allow remote icons in plugins. This is a big privacy issue.

psobolewskiPhD: Actually, this is a really important point, it let's someone track things, like the 1px by 1px white image in your emails.

brisvag: Ok, I agree. I was just trying to keep previous "behaviour" :)

brisvag:
Ok, urls no longer accepted.

TimMonko (commented):
Implementation looks good to me
Can we get some documentation in napari/docs about the "ship with sdist" bit. I think `src/plugin/some_module/icon.svg` is not by default included in the sdist because it is not a .py

This will result in plugin developers _locally_ seeing everything as ok (because their `-e` install contains it) but then the wheel installed version would be wrong. 

But also, maybe we shouldn't reference sdist here? Wheels are always built with the sdist (AFAIK), but that's still managed via `pyproject.toml` not `MANIFEST.in`. ~~So maybe we just link to a good packaging resource that talks about how to include files that are not .py files in builds~~ Ok now I'm a bit confused, has this changed? In ndevio I have .tiff and .png's as sample data but looking closely I did absolutely nothing to include them in the wheel... hmmm, no MANIFEST and nothing in pyproject
Have things changed?

Czaki:
> Ok now I'm a bit confused, has this changed? In ndevio I have .tiff and .png's as sample data but looking closely I did absolutely nothing to include them in the wheel... hmmm, no MANIFEST and nothing in pyproject

It is because you use `hatchling` as build backend, and this that you described is connected with `setuptools` backend. Hatchling includes all data from repository by default.

brisvag:
I update the docstring to clarify the above a bit and fix the bullet points. @Czaki I don't think we can easily have multiline strings here without some extra machinery and having to lock in into a certain formatting for each line as well. I think it's better like this!

TimMonko:
Love the new docstring! I think it clarifies nicely. I'm seeing more plugins use hatch lately, especially those that don't use the plugin-template, so its good to be more agnostic in the language. 

And thanks for looking closely at that Grzegorz. It was my suspicion, especially because I could only find setuptools info on the data inclusion when I was checking the validity of my claim.

TimMonko (approved):

willingc (approved):

src/npe2/manifest/_validators.py
```diff
@@ -67,14 +67,24 @@ def display_name(v: str) -> str:
...
-    if v.startswith("http"):
-        if not v.startswith("https://"):
-            raise ValueError(
-                f"{v} is not a valid icon URL. It must start with 'https://'"
-            )
-        return v
-    assert isinstance(v, str), f"{v} must be a string"
-    return v
+def _ensure_valid_resource(value):
+    if ":" not in value:
```
TimMonko: does this change not invalidate using the fonticon key method? AFAICT, this makes it so only package resources would not raise the value error.

tests/test_manifest.py
```diff
@@ -182,7 +183,18 @@ def test_visibility():
 
 
 def test_icon():
-    PluginManifest(name="myplugin", icon="my_plugin:myicon.png")
+    pm = PluginManifest(name="myplugin", icon="my_plugin:myicon.png")
```
TimMonko, line 187: instead of just testing a package resource, perhaps we should also test against a fonticon key so my above question doesn't happen.

TimMonko (commented):
Was about to merge and wanted to double check the logic because I was trying to update the PR description properly. Looks like maybe we dropped fonticon compatibility? And, I guess I'm not seeing where we block urls now?
I could just be having a brain fart...

brisvag:
Mmm you're right... The problem is that validating the manifest should not necessarily require that the resources are present or the fonticon plugin is isntalled, so it's hard to properly check this... I guess we can check that a `.` is in the text in the same way, but at this point I wonder if checking anything makes sense here...

TimMonko:
I think it makes sense to validate still, especially to intentionally block urls.
Maybe just do `urllib.parse` since its a python builtin? Then yeah, for the resource check just look for .
Although, is there any instance where a uri would work in general? Could just check for `//`? It's not possible for a package resource or fonticon key to have that

brisvag:
So, good news is I updated the validator and now it should handle all the cases... bad news is: we actually have url icons in tests, which made me realize: this is a breaking change in theory... do we need to migrate? Do we need to check if plugins in the wild are setting the icon?

TimMonko:
> Do we need to check if plugins in the wild are setting the icon?

I will check manifests in npe2api. If you don't hear from me by the time you wake up tomorrow send me a DM

TimMonko:
Ok, so I fixed the tests (the sample napari.yaml was still using a url) and I added some more things to `test_icon` because I was curious about submodules and the commands/submenus. 

Checking on the icon usage in npe2api now

TimMonko:
Not a single plugin scraped by npe2api declares an icon except napari-trackastra, and it uses `"icon": "resources/icon.png",`
I used the regex: `"icon"\s*:\s*"(?!")([^"]+)` to find anything that was `"icon": "*"` and it would skip any that are null (no quote) and `""` (quote follows quote) which are the default entries.

There are five other instances that meet this regex, but all are colors used for the three theme plugins that exist: napari-gruvbox, napari-solarized, ndev-themes

So as long as our conda-only plugins don't do this we are good to go. But the conda-only plugin authors seem to be quote responsive these days.

TimMonko (approved):

brisvag:
> resources/icon.png

Hold on, this won't work now! It needs to be `some.module.resources:icon.png`

Maybe we can contact them and open a PR.

Czaki:
What is currrent status. Three plugins that are listed by @TimMonko are @brisvag, @aganders3 and @TimMonko plugin? 

Could we merge this PR and make release, or should we wait?

TimMonko:
> What is currrent status. Three plugins that are listed by @TimMonko are @brisvag, @aganders3 and @TimMonko plugin?
> 
> Could we merge this PR and make release, or should we wait?

The only one of concern is napari-trackastra `"icon": "resources/icon.png", as @brisvag says https://github.com/napari/npe2/pull/450#issuecomment-4572313704

The theme plugins are unaffected by this, I was mostly just showing that I did do diligence to look for "icon" (which themes change but are unaffected by this issue)

I think our best way forward is contact napari-trackastra and merge this PR

brisvag:
I'll do the contacting, feel free to merge!

Czaki (approved):

PR #451, opened 2026-04-04
Fix model schema generation

dstansby:
I noticed this typo over in a failed doc build in `napari`. I guess it was never noticed because it's under a `try: .. except Exception` block.

### DIFF
--- a/_docs/render.py
+++ b/_docs/render.py
@@ -186,7 +186,7 @@ def main(dest: Path = _BUILD):
             schema = json.load(f)
     else:
         try:
-            schema = PluginManifest.model_json_shema()
+            schema = PluginManifest.model_json_schema()
         except Exception:
             with urlopen(SCHEMA_URL) as response:
                 schema = json.load(response)


willingc (approved):
Nice catch @dstansby. Thank you!

willingc:
@TimMonko I would recommend landing this prior to troubleshooting other issues.

willingc:
@TimMonko Consider unpinning Pydantic too.

TimMonko:
Thank you!! I saw this error pop up too when I tried to release npe2 yesterday. This is at least 1 fix :)

> @TimMonko Consider unpinning Pydantic too.

@willingc can you clarify? There is only a lower pin on pydantic, that I don't know the history of. But the exclusion is the just merged #449. Do you want the lower pin to be dropped?

willingc:
@TimMonko This can wait until Monday. I believe that there was a PR to exclude a version of Pydantic. I think that may no longer apply after this change.

PR #453, opened 2026-04-04
Remove render.py from mypy pre-commit exclusion

willingc:
After landing #451, remove `render.py` as an exclusion from pre-commit config.


TimMonko (approved):
Thank you Carol!

PR #454, opened 2026-04-05
Add rationale for pydantic 2.11.x exclusion

willingc:
Thanks to #449 @jasper-tms helped us return our CI to passing. :heart:

Now that bug fix #451 and #453 have landed to fix a render bug, let's see if we can relax this pydantic exclusion. It appears we cannot https://github.com/napari/npe2/pull/454#issuecomment-4189613525
So PR scope updated to reference an issue comment.


jasper-tms:
I quickly tried the same test I wrote in #448 but using this branch of npe2, and I still get the same error fyi! And still specifically with pydantic 2.11

willingc:
Thanks @jasper-tms. I did some digging to figure out why 2.11 is erroring. Keeping 2.11.x excluded and add a comment to link to the rationale for exclusion.

TimMonko:
Thank you for checking @jasper-tms. This is super duper helpful. 😁

TimMonko (approved):

TimMonko (approved):
Updated PR description to new scope. Thanks @willingc !

PR #456, opened 2026-04-10
Remove stale fetch test that has been skipped for over 2 years

willingc:
This PR removes a stale test that uses `napari/dummy-test-plugin`. The test is currently being skipped and has been skipped for almost 3 years.

`napari/dummy-test-plugin` was archived.

We have 2 options to fix this skipped test:
1. Remove the test
2. Find a new example to test instead of `napari/dummy-test-plugin`

This PR uses Option 1 and removes the test.

### DIFF
--- a/tests/test_fetch.py
+++ b/tests/test_fetch.py
@@ -21,16 +21,6 @@ def test_fetch_npe2_manifest():
     assert mf.npe1_shim is False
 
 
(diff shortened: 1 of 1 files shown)


brisvag (approved):
Oof... We can alwasy add something else later, let's remove this :P

PR #461, opened 2026-04-18
Defer importing jsonschema.exceptions

psobolewskiPhD:
Part of: https://github.com/napari/napari/issues/8689

The eager loading of this module costs ~20 ms in napari import time.
This uses PEP 562 to defer the import until it's called. Given that it's behind a try/except anyways, it seems harmless to defer this.

### DIFF
--- a/src/npe2/manifest/contributions/_configuration.py
+++ b/src/npe2/manifest/contributions/_configuration.py
@@ -6,7 +6,6 @@ from ._json_schema import (
     Draft07JsonSchema,
     JsonType,
     JsonTypeArray,
-    ValidationError,
     _coerce_type_name,
 )
 
@@ -116,9 +115,11 @@ class ConfigurationProperty(Draft07JsonSchema):
 
     def validate_instance(self, instance: object) -> dict:
         """Validate an object (instance) against this schema."""
+        from ._json_schema import ValidationError as JsonSchemaValidationError
+
         try:
             return super().validate_instance(instance)
-        except ValidationError as e:
+        except JsonSchemaValidationError as e:
             if e.validator == "pattern" and self.pattern_error_message:
                 e.message = self.pattern_error_message
             raise e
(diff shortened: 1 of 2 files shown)


TimMonko:
Seems reasonable as a solution, but this whole code seems ... odd? Do you know where napari would even be bringing in `jsonschema`? It's not in the core deps and if `Exception` is enough -- why even import it in the first place?

psobolewskiPhD:
It's just cause stuff is eager.
Here's where it is in napari:
image

If you look you can see that `npe2.manifest.contributions` which is used in napari.plugins._npe2.py results in `_configuration` being imported, which imports from _json_schema.py
This PR makes the import of the package `jsonschema.exceptions` deferred so it's not loaded unless it's used.

TimMonko:
Sorry, I wasn't clear. I meant, since npe2 doesn't depend on jsonschema, do we know if it is worth even importing here at all? It's used in testing. Maybe another package transitively brings in the jsonschema. But if its just for testing (here) should we redo the structure instead?

psobolewskiPhD:
I dunno, i've never really done anything with npe2 or plugins. I just traced the import.
In terms of refactoring, maybe @DragaDoncila would have thoughts?
I think we could defer for now, while we consider a refactor? 🤷 
The approach here won't change any actual behavior that uses _json_schema.py.

TimMonko (approved):

DragaDoncila (approved):
@psobolewskiPhD thanks for the ping. I had never seen this part of the codebase before so I had to do some digging. It looks like this file (and other associated code) is defining a `ConfigurationContribution` i.e. a contribution that plugins can declare in the manifest to provide settable attributes/configuration for... something. It's not clear to me if this was supposed to be plugin-wide config, or for specific widgets, or for the viewer....

jsonschema is used here to validate that the given configuration is valid json/pydantic model. 

This code is not used/exposed anywhere -- you can see we [hide the docs](https://github.com/napari/npe2/blob/main/src/npe2/manifest/contributions/_contributions.py#L48-L57) for it, so this PR I think is definitely safe.

In terms of refactoring I'd say our options are:

- figure out what this contribution was meant to do, whether it's still useful, and how we can use it/integrate it into napari
- remove it and its associated code entirely

It'd be nice to figure out what the original intention was behind the Contribution. I can imagine ways in which it could be useful, but I guess none of us really knew about it.

PR #473, opened 2026-06-09
Accept `pathlib.Path` in reading and writing utilities

omkar-334:
The reading and writing utilities accepted only `str` paths. This widens them to also accept `pathlib.Path`, in addition to `str`, so callers (napari included) no longer have to stringify paths before calling npe2.

The approach is to accept `PathLike` (`str | pathlib.Path`) at the public API and normalise `Path` to `str` with `os.fspath` before any path matching or plugin dispatch. Reader and writer plugins keep receiving `str`, so existing plugins are unaffected. Remote URLs are passed as `str` and pass through unchanged.

Changes:

- `types.PathLike` is now `str | pathlib.Path` (was `str`).
- `io_utils`: `read`, `read_get_reader`, `write`, `write_get_writer` (and the internal `_read` / `_write` / `_get_compatible_readers_by_choice`) accept and normalise `Path`.
- `PluginManager`: `iter_compatible_readers` and `get_writer` accept and normalise `Path`.

Resolves the npe2 side of napari/napari#8586.

## Testing strategy

- Added `Path` based tests in `tests/test__io_utils.py` covering `read`, `read_get_reader` (single and stacked), `write`, and `write_get_writer`.
- Added a test asserting that reader plugins still receive `str` when a `Path` is passed, to lock in backwards compatibility.
- Added a `Path` test for `PluginManager.iter_compatible_readers` in `tests/test_plugin_manager.py`.
- Existing `str` based tests are unchanged and still pass, confirming no behaviour change for current callers.

Run:

```
uv run --extra testing pytest -q
```
Screenshot 2026-06-09 at 3 07 36 PM

### DIFF
--- a/src/npe2/io_utils.py
+++ b/src/npe2/io_utils.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 from collections.abc import Sequence
 from typing import (
     TYPE_CHECKING,
@@ -11,7 +12,7 @@ from typing import (
 
 from . import PluginManager
 from .manifest.utils import v1_to_v2
-from .types import FullLayerData, LayerData
+from .types import FullLayerData, LayerData, PathLike
 
 if TYPE_CHECKING:
     import napari.layers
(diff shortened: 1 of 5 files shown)


src/npe2/io_utils.py
TimMonko: this assertion is now too tight if we switch the typing from `list` to `Sequence`. In addition, `_read()` will normalize the Sequence -> list[str], so we don't need to check if its a list here (that's _read's job now)
```suggestion
```

TimMonko: Maybe you could move the `_normalize_paths()` call to this function rather than in `_read`? I think it's ok in either place though

src/npe2/io_utils.py
```diff
@@ -11,22 +12,35 @@
...
+    """Normalise ``PathLike`` input to ``str`` for downstream/plugin use.
+
+    npe2 accepts ``str`` or ``pathlib.Path`` at its public API, but reader and
+    writer plugins (and the internal matching logic) expect ``str``. ``Path``
+    objects are converted with ``os.fspath``; ``str`` (including remote URLs)
+    is returned unchanged.
+    """
+    if isinstance(paths, (str, os.PathLike)):
+        return os.fspath(paths)
+    return [os.fspath(p) for p in paths]
```
TimMonko, line 33: great! and this is a very helpful docstring in the long run, because it clearly explains why we shouldn't pass `Path`s downstream

TimMonko (commented):
@omkar-334 thank you very much for your two contributions (also napari/napari#9042). These look to be in pretty great shape so far. I only have one comment for this PR :)

I appreciate especially the helper function and its docstring you've added. Your test coverage is also clear and thorough and should greatly help with unintentional regressions, so I appreciate that too! 

Welcome to our project!! I'm sorry it took me a bit to get around to reviewing this -- I knew it would take a bit for me to sit down and think about and we were quite busy prepping for release this past week. I appreciate your patience and I look forward to working through these with you!

omkar-334:
hey @TimMonko happy to be here!  
thanks for reviewing these two PRs -> I have made the requested changes on both of them. The `mypy` errors were a bit confusing but your suggestions made sense.

TimMonko (approved):
Thanks, looks good now! I'm going to merge this and make towards an npe2 release, to make progress on napari/napari#9042

It will likely be able to be fully released next week after our pre-release checks run (since I'm headed out on vacation soon). Just wanted to give you an update :) 🍻

PR #377, opened 2025-05-10
Improve reading error for selected plugins

DragaDoncila:
When a plugin is **selected** by the user to read a file, we want to raise specific error messages when that reader fails to read, for whatever reason.

Prior to this PR:

- if a selected reader refused to read the file i.e. the `get_reader` function returned `None`, the error message would say that the plugin `returned no data`
- if a selected reader returned a `null_layer_sentinel`, **no error would be surfaced in `napari`**. I think this is a holdover from olden days when we would cascade readers, so we didn't want the `null_layer_sentinel` to throw an error. I think there's probably good reason to remove the `null_layer_sentinel` entirely, but that's a discussion for another day.

With this PR:

- if a selected reader refused to read the file, we get `"Reader {plugin_name!r} was selected to open {paths!r}, but refused the file."`
- if a selected reader tried to read the file and returned `null_layer_sentinel`, we get `"Reader {plugin_name!r} was selected to open {paths!r}, but returned no data."`

As an aside, if a plugin is **not selected**, we get a semi-generic `"No readers returned data"` but I think this is also ok as, from the `napari` side, it's ~impossible for npe2 reading to get called without a plugin selected - either we use the preferred one, the user specifically selects one, or we explicitly pass the only one available.

Note that this brings the `_is_null_layer_sentinel` function to `npe2`, and it previously lived in `napari`. I think this is ok, and if we're happy with this PR, we can update its usage in `napari` to import from here.

Note also that this function checks specifically for a `list` (same as with napari/#7851), not a container sequence, but I actually think this is also ok and maybe we should update our typing in other places to simply expect a list, since many other container sequences wouldn't be useful anyway: a `dict` would likely fail somewhere weirdly, a `set` would lead to unexpected layer orderings... I guess a generator would be ok? But I don't think anyone is returning anything except a list.

PR #474, opened 2026-06-15
Log Warning for null layer sentinel instead of ValueError

TimMonko:
# References and Relevant Issues

follow-up to #377
discussion on zulip starting here: https://napari.zulipchat.com/#narrow/channel/215289-release/topic/0.2E7.2E1/near/603315795

# Description

In #377 this moved the _is_null_layer_sentinel check to npe2.
Originally introduced in napari/napari#2206, the null layer sentinel `[(None,)]` has since then been used by plugins to signal that they have successfully processed the file but have no layer data to return. 
#377 broke part of this contract, because the check now resulted in raising a ValueError. This means that the code prior to the null_layer_sentinel would still run (like the builtin script reader or a plugin which opens a widget), but the ValueError would be raised and break tests -- causing a regression for both our script reader https://github.com/napari/napari/actions/runs/27539941390 and for plugins using the sentinel.

The goal of this check in #377 was to better inform the user about what happens, so I've opted to use the warnings.warn() machinery (from double checking against my own napari/napari#8914 notifications example) -- this will still inform the users that the sentinel was returned, but won't break plugins.

I also added lots of comments in the code because the mental flow was difficult for me, (and I did reorder the logic a bit) -- but if we don't want comments just use the first commit. 

Originallly, this was implemented as a UserWarning, but has since been moved to a logger.DEBUG

### DIFF
--- a/src/npe2/io_utils.py
+++ b/src/npe2/io_utils.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import warnings
+import logging
 from collections.abc import Sequence
 from typing import (
     TYPE_CHECKING,
@@ -19,6 +19,7 @@ if TYPE_CHECKING:
 
     from .manifest.contributions import ReaderContribution, WriterContribution
 
+logger = logging.getLogger(__name__)
 
 def read(
     paths: list[str], *, stack: bool, plugin_name: str | None = None
@@ -181,14 +182,11 @@ def _read(
         # used e.g. by napari's built-in .py reader (scripts modify the
         # viewer directly) and by third-party plugins that handle files
         # through non-layer pathways, like opening a widget.
-        # We warn informatively when the user explicitly chose this plugin,
-        # but still return the data so the caller can handle it.
         if plugin_name and _is_null_layer_sentinel(layer_data):
-            warnings.warn(
+            logger.warning(
                 f"Reader {plugin_name!r} was selected to open {paths!r}, "
-                "but opened no layers`. This may be intentional. If you "
-                "expected layers to be opened, contact the plugin author",
-                stacklevel=2,
+                "and opened no layers. This may be intentional. If you "
+                "expected layers to be opened, contact the plugin author."
             )
 
         # ── reader returns truthy ────────────────────
--- a/src/npe2/io_utils.py
+++ b/src/npe2/io_utils.py
@@ -21,6 +21,7 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+
 def read(
     paths: list[str], *, stack: bool, plugin_name: str | None = None
 ) -> list[LayerData]:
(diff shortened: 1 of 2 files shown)


src/npe2/io_utils.py
```diff
@@ -163,28 +164,58 @@ def _read(
...
+        # has no layer data to return. This is a documented contract value
+        # used e.g. by napari's built-in .py reader (scripts modify the
+        # viewer directly) and by third-party plugins that handle files
+        # through non-layer pathways, like opening a widget.
+        # We warn informatively when the user explicitly chose this plugin,
+        # but still return the data so the caller can handle it.
+        if plugin_name and _is_null_layer_sentinel(layer_data):
+            warnings.warn(
+                f"Reader {plugin_name!r} was selected to open {paths!r}, "
+                "but returned the null layer sentinel `[(None,)]`. This "
```
DragaDoncila:
```suggestion
                "but opened no layers`. This "
```

src/npe2/io_utils.py
```diff
@@ -163,28 +164,58 @@ def _read(
...
+                "may be intentional — e.g. the reader may have executed "
```
DragaDoncila: Not sure I love blessing this code pathway via error message. I would say "this may be intentional. If you expected layers to be opened, contact plugin author" or something like that.

Czaki:
Might we focus on ship blessed way. Then we could add a proper warning?

TimMonko:
> Might we focus on ship blessed way. Then we could add a proper warning?

@Czaki , I'm not sure I understand what you are saying. This PR (or something else to remove the `ValueError`) is important to unblock an npe2 release.
In terms of a blessed pathway the `null_layer_sentinel` has been and continues to be blessed, its the arbitrary code execution before returning the null layer sentinel that is not blessed. 
So, this logs a warning (and unblocks release) while we figure out a proper implementation.
I could be misunderstanding your intent though

Czaki:
I think that we could add a blessed way. We could add a filed to reader declaration. For example:

```yaml
  - command: example-plugin.read_xyz
    filename_patterns:
    - '*.xyz'
    need_viewer: True
```

For a reader, that will require viewer for work and is guaranteed to be executed in main thread? Then we could emit warning for a reader that does not have such attribute and return sentinel?

TimMonko:
> I think that we could add a blessed way. We could add a filed to reader declaration. For example:
> 
> ```yaml
>   - command: example-plugin.read_xyz
>     filename_patterns:
>     - '*.xyz'
>     need_viewer: True
> ```
> 
> For a reader, that will require viewer for work and is guaranteed to be executed in main thread? Then we could emit warning for a reader that does not have such attribute and return sentinel?

Yeah! I love these kind of ideas. But, is there any reason to not move forward with this PR in the meantime? I want to unblock release of npe2 😆

Czaki (approved):

PR #484, opened 2026-07-18
Select writers by compound file extensions (e.g. `.ome.tiff`, `.ome.zarr`)

### LINKED ISSUE #9088: [plugin, save dialog] ome.tiff matches with .tiff of builtins
From: https://github.com/napari/napari-tiff/pull/62/changes#r3460473613

Not sure if this is intended, but normally when you use the dropdown to change the writer, the etension in the text entry field updates.
But the default writer is builtins .tiff so if a plugin uses ome.tiff, it doesn't update. This means someone could over-write a .tiff when wanting to save an ome.tiff. As a result I think this is a bug?

There may be other situations this could happen, but ome.tiff scenario is probably the most common (zarr/ome.zarr is another). It could affect other plugins that want to write this common format. 

Note: this also happens with reading (see: https://github.com/napari/napari-tiff/pull/89#issuecomment-4762477958) where a reader declaring `*.tiff` will accept `*.ome.tiff`. Again not sure if intentional, but it's probably safe there since the worst case is a pure tiff reader losing the metadata?

omkar-334:
earlier, `PluginManager.get_writer` chose a writer by the final suffix only, via `Path(path).suffix`. For a path like `foo.ome.tiff` that suffix is just `.tiff`, so a writer registering `.ome.tiff` was skipped and the plain `.tiff` writer was picked instead. The same happened for `.ome.zarr` versus `.zarr`. In the save flow this means a file the user meant to save as OME-TIFF could be handed to the plain TIFF writer.

# Fix i implemented
The fix matches a writer when the filename ends with one of its declared extensions, and prefers the longest match, so `.ome.tiff` wins over `.tiff`. Ties keep the existing writer priority (iteration order). Paths without an extension are unchanged: a single layer still gets the writer's default extension appended, and multi layer saves still fall back to a directory writer.

## Testing

- Added `test_get_writer_compound_extension` in `tests/test__io_utils.py`. It registers a plain `.tif`/`.tiff` writer first and an `.ome.tif`/`.ome.tiff` writer second, then asserts that a compound path still selects the OME writer, a plain path selects the plain writer, and a path with no extension appends the default.
- Existing writer tests are unchanged and still pass.

Run:

```
uv run --extra testing pytest tests/test__io_utils.py -q
```

Resolves https://github.com/napari/napari/issues/9088

### DIFF
--- a/src/npe2/_plugin_manager.py
+++ b/src/npe2/_plugin_manager.py
@@ -698,7 +698,7 @@ class PluginManager:
         path = os.fspath(path) if path else ""
         # Match against the whole filename rather than `Path.suffix`, so a
         # compound extension like ".ome.tiff" is not shadowed by a ".tiff" writer.
-        name = os.path.basename(path).lower()
+        filename = os.path.basename(path).lower()
         has_ext = bool(Path(path).suffix)
 
         if has_ext:
@@ -710,7 +710,9 @@ class PluginManager:
                 if plugin_name and not writer.command.startswith(plugin_name):
                     continue
                 matched = [
-                    e for e in writer.filename_extensions if name.endswith(e.lower())
+                    e
+                    for e in writer.filename_extensions
+                    if filename.endswith(e.lower())
                 ]
                 if matched and (longest := max(len(e) for e in matched)) > best_len:
                     best_writer, best_len = writer, longest
--- a/src/npe2/_plugin_manager.py
+++ b/src/npe2/_plugin_manager.py
@@ -696,19 +696,35 @@ class PluginManager:
         """
         # normalise Path -> str for suffix matching and `path + ext` below
         path = os.fspath(path) if path else ""
-        ext = Path(path).suffix.lower() if path else ""
+        # Match against the whole filename rather than `Path.suffix`, so a
+        # compound extension like ".ome.tiff" is not shadowed by a ".tiff" writer.
+        name = os.path.basename(path).lower()
+        has_ext = bool(Path(path).suffix)
 
+        if has_ext:
+            # Prefer the longest matching extension; break ties by writer
+            # priority (iteration order).
+            best_writer = None
+            best_len = -1
+            for writer in self.iter_compatible_writers(layer_types):
+                if plugin_name and not writer.command.startswith(plugin_name):
+                    continue
+                matched = [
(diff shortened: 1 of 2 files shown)


omkar-334:
Before:  
image  

After:  
image

omkar-334:
The CI is failing because of a new release for the `typer` package - https://github.com/fastapi/typer/releases.  
i raised the PR to fix this - https://github.com/napari/npe2/pull/485

omkar-334:
hey, the tests are now passing since we merged the PRs, and i've fixed the mypr/codecov errors as well

src/npe2/_plugin_manager.py
```diff
@@ -696,19 +696,35 @@ def get_writer(
         """
         # normalise Path -> str for suffix matching and `path + ext` below
         path = os.fspath(path) if path else ""
-        ext = Path(path).suffix.lower() if path else ""
-
+        # Match against the whole filename rather than `Path.suffix`, so a
+        # compound extension like ".ome.tiff" is not shadowed by a ".tiff" writer.
+        name = os.path.basename(path).lower()
```
DragaDoncila: Name gets a bit ambiguous below where we're looking through plugin name/filename extensions etc. Let's be explicit here.

```suggestion
        filename = os.path.basename(path).lower()
```

DragaDoncila (approved):
@omkar-334 thanks for the PR! Agreed this is a strict improvement and fixing an existing bug with this function, so let's get this in. 

Unfortunately, we don't actually use this function in the saving workflow, so it won't help with napari/napari#9088 where Peter correctly has identified that the appended extension in the native dialogs isn't under our control. I'm having a think about how we can improve that.

src/npe2/_plugin_manager.py
omkar-334: understood, i've made this change.

omkar-334:
> Unfortunately, we don't actually use this function in the saving workflow, so it won't help with [napari/napari#9088](https://github.com/napari/napari/issues/9088) where Peter correctly has identified that the appended extension in the native dialogs isn't under our control. I'm having a think about how we can improve that.

Understood, thanks. My first thought was that this could be fixed here in npe2, but your and Peter's comments make it clear the mismatch comes from the native save dialog on the napari side

PR #490, opened 2026-07-25
Remove title from configuration contribution

BadPrograms:
Configuration settings should not have a title in the manifest, as it should inherit the title of the whole plugin


DragaDoncila:
We realized that likely, we WOULD want a plugin to be able to declare multiple Configuration contributions, to separate settings for different widgets/readers/writers/etc. on their preference page. If we have multiple, we will want a title, so we're closing this PR for now.

PR #491, opened 2026-07-27
Unify configuration schema and greatly simplify fields

DragaDoncila:
This PR:

- Removes unused Draft04 and Draft06 JsonSchema models
- Renames Draft07 JsonSchema to simply "ConfigurationJsonSchema", and links to the most recent version of the schema, 2020-12. I confirmed we are a compatible subset of that schema format
- Removes a bunch of configuration property fields that are unlikely to be useful for plugin settings:
   - Object-type properties and associated constraint fields
   - Array-type properties and associated constraint fields
   - Properties related to composable schemas (not, any_of, etc.)
   - None-type properties (typically used in composed schemas)
   - Const-type properties (typically used in composed schemas)
 - Removes some constraint fields from property types we did keep:
   - We won't support pattern-validation for string types
   - We won't support format-validation for string types (emails etc.)
 - Removes unnecessary ConfigurationProperty fields -- mostly fields that allow the user to specify whether some text is markdown or plain formatted

What didn't you remove?!

- Boolean, enum, string, integer and float-type configuration properties
- Constraints on integer/float-type properties (min/max/exclusive_min/exclusive_max/multiple_of)
- Constraining a string to a min/max length
- Setting a deprecation_message for a property

### DIFF
--- a/src/npe2/manifest/contributions/_configuration.py
+++ b/src/npe2/manifest/contributions/_configuration.py
@@ -5,7 +5,7 @@ from pydantic import BaseModel, BeforeValidator, Field, conlist, model_validator
 from ._json_schema import (
     ConfigurationJsonSchema,
     JsonType,
-    _coerce_type_name,
+    _to_json_type,
 )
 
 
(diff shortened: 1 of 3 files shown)


brisvag (commented):
Discussed in person: the removal of all those `not_` and `all_of` logical fields actually has the downside that we cannot (for example) declare a field to be an "integer not equal to zero". Alternatively, we need to add a `not_equal` custom field and write our own validation for it. We conlcuded that for now we can just ignore this and deal with it at a later point, and for now such restriction is not allowed.

---

General question: how do "nullable" fields behave? As a plugin developer, I think I should be able to opt in for a field to be nullable. By default I want my `int` field to *always* have an int value, unless I specifically say somehwere that `None` is acceptable.

DragaDoncila:
@brisvag I opened a follow-up issue based on your questions/comments. The latest commit disallowed assigning multiple types to a single property.

brisvag (approved):

TimMonko (approved):

PR #495, opened 2026-07-28
Fully document all `ConfigurationJsonSchema` and `ConfigurationProperty` fields

### LINKED ISSUE #494: All ConfigurationContribution fields need to be documented
Follow-up from #491 - we need to document each individual `Field` for plugin authors to have an easy and complete reference of what everything means w.r.t napari.

@TimMonko and I discussed and decided to document all the 	`ConfigurationJsonSchema` fields directly on the model (rather than duplicating and re-exposing them in `ConfigurationProperty`), and any additional fields that do not match the schema will be stored in and documented directly on `ConfigurationProperty`

DragaDoncila:
Closes #494 

This PR adds a description to all fields in the class they are declared. We no longer re-define `type`, `default`, `description` and `enum` in `ConfigurationProperty`, as they are already defined (and now documented) in `ConfigurationJsonSchema`. 

After this PR, `ConfigurationProperty` only defines **additional** fields that are **not** present in `ConfigurationJsonSchema` and **not** compatible with the JSON Schema spec.

### DIFF
--- a/src/npe2/manifest/contributions/_configuration.py
+++ b/src/npe2/manifest/contributions/_configuration.py
@@ -1,44 +1,16 @@
-from typing import Annotated, Any
+from pydantic import BaseModel, Field, model_validator
 
-from pydantic import BaseModel, BeforeValidator, Field, conlist, model_validator
-
-from ._json_schema import (
-    ConfigurationJsonSchema,
-    JsonType,
-    _to_json_type,
-)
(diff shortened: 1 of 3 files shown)


Czaki (approved):

DragaDoncila:
Got verbal approval from @brisvag at hackathon as well, so will go ahead and merge. Can always improve the docs later!

PR #496, opened 2026-07-28
Make title, type and default required fields for ConfigurationProperty

DragaDoncila:
Previously, only `type` was required for `ConfigurationProperty`. We think, for use in napari settings, `title` and `default` should also be required.

This PR updates the `ConfigurationProperty` and tests accordingly.


TimMonko:
🥳 LGTM. Absolutely agree. zen of Python

TimMonko (approved):

PR #502, opened 2026-08-06
Docs: ConfigurationContribution guide

TimMonko:
# References and Relevant Issues

napari/napari#9308 for the GUI implementation
napari/docs#1100 for the stubgen (I dunno which is needed first? but napari-docs builds from npe2 latest tagged release)

# Description

This adds documentation to discuss how configurations/settings are defined in the napari manifest. It includes information on how the naming is decided, what is required, and how that shows up in both the napari Preferences GUI and is accessed programmatically.

If the document doesn't otherwise speak for itself without more description here, it needs updated.

snippet of Contributions Reference showing both `ConfigurationsContribution` and `ConfigurationProperty`
image

snippet of Guides Page showing jinja inclusion of example manifest (used in the guide)
image

### DIFF
--- a/_docs/templates/_npe2_configuration_guide.md.jinja
+++ b/_docs/templates/_npe2_configuration_guide.md.jinja
@@ -53,6 +53,35 @@ The category title is converted to snake_case to form the attribute name
 each key becomes a snake_case field name on that model (with the plugin-name
 prefix stripped).
 
+### Events: Reacting to settings changes
+
+Settings are napari `EventedModel`s, so changing a value emits a
+[psygnal](https://github.com/pyapp-kit/psygnal) event that you can connect to.
+Each category exposes an `events` emitter group with one emitter per setting:
+
+```python
+my_settings = get_plugin_settings("example-plugin")
+
+def on_max_size_changed(event):
+    print("max size is now", event.value)
+
+my_settings.reader.events.max_size_mb.connect(on_max_size_changed)
+
+my_settings.reader.max_size_mb = 1024  # event causes a print: "max size is now 1024"
+```
+
+The top-level settings object also re-emits every change as a single
+`changed` event carrying a dotted `<category>.<property>` key, so one
+connection covers all of a plugin's settings:
+
+```python
+my_settings.events.changed.connect(lambda e: print(e.key, e.value))
+# after the assignment above -> "reader.max_size_mb 1024"
+```
+
+These events fire whether a value is changed in code or in the Preferences
+dialog, and the change is auto-saved to the config file.
+
 ### Supported types and constraints
 
 Each property follows a subset of
--- a/_docs/templates/_npe2_configuration_guide.md.jinja
+++ b/_docs/templates/_npe2_configuration_guide.md.jinja
@@ -53,6 +53,31 @@ The category title is converted to snake_case to form the attribute name
 each key becomes a snake_case field name on that model (with the plugin-name
 prefix stripped).
 
+### How settings render in napari
+
+For the example plugin above, the Preferences dialog shows one page per
+plugin, with each category rendered as its own group:
+
+```text
+Preferences
+  ...              ── existing napari sections ──
+  ----Plugin Preferences----
+  Example Plugin   ◄── page per plugin (display_name)
+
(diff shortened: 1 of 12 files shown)


_docs/templates/_npe2_configurations_guide.md.jinja
```diff
@@ -0,0 +1,179 @@
...
+If you only read one or two settings, inline the fallback instead:
+
+```python
+def max_size_mb() -> float:
+    try:
+        from napari.settings import get_plugin_settings
+        return get_plugin_settings("example-plugin").reader.max_size_mb
+    except (ImportError, KeyError):
+        return 512.0
+```
```
BadPrograms, line 219: Just want to mention that the event transmission is missing over here, which is one of the most important things in this feature
```python
get_plugin_settings()["demo-napari"].demo_napari.events.some_property.connect(lambda x:print(x))
get_plugin_settings()["demo-napari"].demo_napari.some_property = True
```

TimMonko: Thank you! This is a great idea and I've now pushed such changes to describe it :)

_docs/templates/_npe2_configuration_guide.md.jinja
```diff
@@ -108,3 +109,32 @@ npe2 validates that configuration **titles** and **property keys** are unique
...
+
+    pm = PluginManager.instance()
+    if 'example-plugin' not in pm:
+        pm.register(PluginManifest.from_distribution('example-plugin'))
+    # Reset the in-memory cache so the registered plugin is included (saved
+    # values are re-read from disk; this also allows path_dir=tmp_path).
+    napari_settings._PLUGIN_PREFERENCES.clear()
+
+    s = napari_settings.get_plugin_settings('example-plugin', path_dir=tmp_path)
+    assert s.reader.max_size_mb == 512
```
BadPrograms: Is this a sample attibute or it should exist in all PluginPreferences?

_docs/templates/_npe2_configuration_guide.md.jinja
```diff
@@ -8,7 +8,8 @@ no Python code is needed to build the settings UI.
 
 Configuration contribution values are exposed at runtime via
 `napari.settings.get_plugin_settings("example-plugin")` and persisted to
-`<config dir>/example-plugin.yaml`.
+`<config dir>/example-plugin.yaml`. Settings are written to disk
+*automatically* whenever a value changes.
```
BadPrograms: I'm not sure if it should be mentioned here that each plugin has its own independent file or not.

BadPrograms (commented):
Very nice documentation!

_docs/templates/_npe2_configuration_guide.md.jinja
BadPrograms: Scratch that just saw the yaml :)

TimMonko: yeah its sneaky that it gets built in with the jinja template. I do hope it's sufficient for folks to get the jist that I'm using it across the whole file

Czaki:
I downloaded build docs, and the configuration is level above all other section 
image

I think that it is some bug and should be fixed.

TimMonko:
> I think that it is some bug and should be fixed.

this isn't what I see locally, but also it helped me learn that I didn't push the recent commit I had...

I'll let it rebuild and check out the artificat, but I'm not sure the docs are building correctly generally, given the apperance of the search on your thing

TimMonko:
Ok, I'm looking into what CI does here, but this PR doesn't even touch the TOC of npe2 docs, and I'm not quite sure what CI docs here is trying to accomplish. In napari we see it correctly as:
image

These docs are a jupyter book and doesn't really have anything to do with what we need.
It's because CI does this:
```
          cd docs
          jupyter-book toc from-project . -f jb-book > _toc.yml
          jupyter-book build .
```
Which means it infers the TOC for its build from the folder, and now "Configurations" comes first alphabetically. If we don't want an auto toc and such than that should be a different PR.

DragaDoncila (approved):
Let's get this in so we can get docs building in full. If there's anything misconfigured here we can follow-up.

PR #503, opened 2026-08-11
Make manifest.configuration a dictionary

DragaDoncila:
As discussed in triage meeting today, and addressing concerns raised in [this zulip chat]([#2026-naPLari-Krakow-public > Plugin Settings (ConfigurationContribution) @ 💬](https://napari.zulipchat.com/#narrow/channel/619705-2026-naPLari-Krakow-public/topic/Plugin.20Settings.20.28ConfigurationContribution.29/near/615071514)), this PR updates the manifest `configuration` field to be a dictionary, with each `ConfigurationContribution` being given a unique key within the manifest.

This allows us to use the key on the napari side to construct the key used with `get_plugin_settings()`.

Prior to this PR, the title of each `ConfigurationContribution` would be "keyified" on the napari side, and concatenated with a stripped version of the `ConfigurationProperty` key for programmatic access. This made it more difficult for plugin developers to reason about how to programmatically access each property, and made it harder to define the necessary "uniqueness" of each field.

Now, each `ConfigurationContribution` key must be unique within a manifest, and each `ConfigurationProperty` key must be unique within its `ConfigurationContribution`. Both keys are explicitly declared in the manifest, and titles are used only for display.

Note this means we undo the validation added in #501, because the only validation needed is that the keys are valid Python identifiers, and unique*.

The only other changes in this PR are:

- rename manifest.configuration to manifest.configurations to match all our other fields and indicate it can hold multiple
- bump the schema version (since technically any previously written manifests would now be invalid)

* See #504 for caveat on this

### DIFF
--- a/src/npe2/manifest/contributions/_configuration.py
+++ b/src/npe2/manifest/contributions/_configuration.py
@@ -1,61 +1,10 @@
-import re
-from collections.abc import Iterable
+from typing import Annotated
 
-from pydantic import BaseModel, Field, model_validator
+from pydantic import AfterValidator, BaseModel, Field, model_validator
+
+from npe2.manifest import _validators
 
 from ._json_schema import ConfigurationJsonSchema
-
-
-def normalize_title(title: str) -> str:
-    """Return a canonical snake_case key for a configuration title.
-
-    Configuration titles are free-form text (e.g. "Demo Configuration for
-    widget 1") but consumers (e.g. napari) use them to derive identifier-like
-    keys / field names.  Normalizing titles to a canonical key lets npe2
-    enforce that titles are unique within a plugin, and lets consumers reuse
-    the same normalization instead of implementing (possibly divergent) ones.
-
-    Examples
-    --------
-    >>> normalize_title('Demo Configuration for widget 1')
-    'demo_configuration_for_widget_1'
-    >>> normalize_title('main widget')
-    'main_widget'
-    """
-    # camelCase -> snake_case (e.g. someSetting -> some_Setting)
-    name = re.sub(r"(?<!^)(?=[A-Z])", "_", title)
-    # any run of non-alphanumeric characters is a separator
-    name = re.sub(r"[^0-9a-zA-Z]+", "_", name)
-    name = re.sub(r"_+", "_", name).strip("_").lower()
-    if not name:
-        name = "settings"
-    if name[0].isdigit():
-        name = f"_{name}"
-    return name
-
-
-def _ensure_unique_normalized(items: Iterable[str], what: str) -> None:
-    """Raise if any two strings in ``items`` normalize to the same key.
-
-    Parameters
-    ----------
-    items : Iterable[str]
-        The free-form strings to check for collisions after
-        :func:`normalize_title`.
-    what : str
-        Human-readable noun phrase used in the error message, e.g.
-        "Configuration titles".
-    """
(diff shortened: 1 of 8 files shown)


DragaDoncila:
@TimMonko for your review. Also note @BadPrograms this is going to need changes on the napari PR side, mainly:

- use `.configurations` everywhere
- update how we iterate through `.configurations` as it's now a dict
- use the configuration key rather than title when constructing the models
- I think we can then get rid of the snakecase thing

src/npe2/manifest/contributions/_configuration.py
```diff
@@ -102,34 +51,40 @@ def _validate_root(cls, values):
...
-        "breaks. For example, if your key is 'gitMagic.blame.dateFormat', the "
-        "generated title for the setting will look like 'Blame: Date Format'",
+        description="Configuration properties, keyed by a property key that is local "
+        "to this configuration contribution. Each property key must be a valid, "
+        "non-reserved Python identifier that does not begin with an underscore, "
+        "since it is used verbatim as the attribute name for that setting on the "
+        "generated settings model (e.g. "
+        "`get_plugin_settings('plugin-name').<configuration-key>.<property-key>`). "
+        "Property keys only need to be unique within this configuration "
+        "contribution, not across the whole plugin.",
```
TimMonko, line 86: this is important, but your PR description reads like it has to be unique plugin-wide.

src/npe2/manifest/contributions/_contributions.py
```diff
@@ -48,27 +45,16 @@ class ContributionPoints(BaseModel):
...
-        "settings. If there are multiple categories of settings, the Settings editor"
-        "will show a submenu in the table of contents for that extension, and the title"
-        "keys will be used for the submenu entry names.",
+        description="Configuration options for this plugin, keyed by a configuration "
+        "key. A plugin can contribute multiple categories of settings by declaring "
+        "multiple entries here; each shows up as its own submenu in the Settings "
+        "editor, headed by that entry's `title`. Each configuration key must be a "
+        "valid, non-reserved Python identifier that does not begin with an "
+        "underscore, and must be unique within the manifest, since it is used "
+        "verbatim as the attribute name for that settings category on the generated "
```
TimMonko:
```suggestion
        "verbatim as the attribute name for that setting's category on the generated "
```

tests/sample/my_plugin/napari.yaml
TimMonko, line 1: absolutely this is way better :)

TimMonko (approved):
> Note this means we undo the validation added in #501, because the only validation needed is that the keys are valid Python identifiers, and unique*.

Ah, good to see! I know I mentioned this in my quick #505 comment. I think this removal is indicative of a good thing, insofar that we remove special casing and just get it back to reasonable declarative roots. 

Done with my review now and it's beautiful 🤩, 10/10. This makes me feel so much better about presenting this in a solid and maintainable form. Great work, thanks for participating in this discussion @carlosmariorr !

src/npe2/manifest/contributions/_configuration.py
DragaDoncila: @TimMonko can you suggest a change (or directly edit) the PR description? I'm not sure which bit suggests that the property key needs to be plugin-wide

TimMonko: me misreading, I think

PR #505, opened 2026-08-11
Disallow duplicate keys in the same yaml/json mapping

### LINKED ISSUE #504: Duplicate manifest keys in YAML or JSON files get silently dropped
When we define a dictionary in yaml or json manifests e.g. `manifest.contributions.menus` or (soon) `manifest.contributions.configuration`, the parser doesn't complain when duplicate keys are declared, but simply overwrites any existing values at that key with the last one it reads.

We need to fix this in the parser.

DragaDoncila:
Closes #504 

This PR adds a yaml and json parser that ensure duplicate keys are not declared in the same mapping e.g. in the `menus` contribution dict, the `configurations` contribution dict (being added in #503), or within a single `ConfigurationContribution` when keying individual properties.

TOML already disallows this, but added a regression test for the sake of completeness.

This PR was heavily Claude assisted, but I've reviewed it and it seems to make sense to me!

### DIFF
--- a/src/npe2/manifest/_bases.py
+++ b/src/npe2/manifest/_bases.py
@@ -7,6 +7,53 @@ import yaml
 from pydantic import BaseModel, PrivateAttr
 
 
+class _UniqueKeyYamlLoader(yaml.SafeLoader):
+    """YAML loader that raises on duplicate keys within a single mapping.
+
+    PyYAML (like the stdlib `json` module) silently keeps the *last* value
+    for a duplicate key within a mapping, which would let e.g. two
+    `contributions.configurations` entries share a key and silently
+    discard one during parsing -- before pydantic ever sees the data to
+    validate it. This loader catches that at parse time instead.
+    """
+
+    def construct_mapping(self, node, deep=False):
+        seen = set()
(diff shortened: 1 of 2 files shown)


DragaDoncila:
Leaving this in draft in case it's like, overkill, or I'm doing something weird. @Czaki wanna take a look?

TimMonko:
I agree with this in principle. Not doing code review rn. I thought about going this route in #501 <- we should make sure that's up to date now given #503

TimMonko (approved):
Did code review now-- and as I mentioned this was something I already had in mind. It would be nice to have a test here that passes in a problematic manifest and checks for the errors.

DragaDoncila:
>  It would be nice to have a test here that passes in a problematic manifest and checks for the errors.

Thanks for flagging, made me realize I didn't push the tests :sweat_smile: Pushed now

TimMonko:
I think one possibility where this could bite someone is in the instance where they might be copy-pasting template or other working contributions, but not updating the previous names. In that case they would have repeats in the manifest that were silently squished but it _worked_ for them, unknowingly why, because they only cared about the last contribution.

PR #506, opened 2026-08-13
Comment unimplemented ConfigurationProperty fields

DragaDoncila:
We won't get a chance to implement behaviour for these on the napari side ahead of the 0.9.0 release, so we're commenting them out for now (so they don't get documented).

We'll improve for 0.9.1.


TimMonko (approved):

TimMonko:
I can update tests tomorrow if needed

PR #508, opened 2026-08-18
Validate Enum values for ConfigurationJsonSchema

TimMonko:
# Description

In napari/napari#9308, I discovered that enum values could be anything arbitrary, and fixed it. This upstreams Enum member validation to npe2, so that, at least, plugin authors don't hit it prior to testing in napari. It maps Enum member types to JSON schema types, and checks that the Enum members are of the correct type for the JSON schema type. It also makes sure that the default value is an Enum member.

### DIFF
--- a/src/npe2/manifest/contributions/_json_schema.py
+++ b/src/npe2/manifest/contributions/_json_schema.py
@@ -84,6 +84,21 @@ _python_equivalent: dict[str, type] = {
 }
 
 
+def _enum_member_type_ok(member: Any, json_type: str) -> bool:
+    """Return True if ``member`` is an acceptable value for ``json_type``."""
+    # ``bool`` is a subclass of ``int`` in Python, so exclude booleans from
+    # the numeric types explicitly.
+    if json_type == "boolean":
+        return isinstance(member, bool)
+    if json_type == "integer":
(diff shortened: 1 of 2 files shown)


src/npe2/manifest/contributions/_json_schema.py
```diff
@@ -84,6 +84,21 @@ def _to_camel_validation(string: str) -> AliasChoices:
...
+    # the numeric types explicitly.
+    if json_type == "boolean":
+        return isinstance(member, bool)
+    if json_type == "integer":
+        return isinstance(member, int) and not isinstance(member, bool)
+    if json_type == "number":
+        return isinstance(member, (int, float)) and not isinstance(member, bool)
+    if json_type == "string":
+        return isinstance(member, str)
+    return False
```
DragaDoncila, line 99: This line isn't covered which I guess happens when you don't pass in a valid json_type? What gets raised in that case? Is it the "enum members" error? Cause I think it should actually be a different error in that case right? About how the type is unrecognised?

DragaDoncila (approved):
Looks ok to me - ideally we'd fix the coverage because it should be easy to do and it makes sense to maintain the 100%. Even if we don't fix coverage would appreciate an answer to the question for my own edification.

PR #512, opened 2026-09-18
Add default dock area field

brisvag:
Allows setting a `default_area` for a widget. We expose this machinery in napari for manually added widgets, but not for plugins.

Companion PR: https://github.com/napari/napari/pull/9526


TimMonko (approved):
LGTM. Should the field name be perhaps "dock_area"? I'm open to either. I thought perhaps even "area" at first, but it's not descriptive enough

brisvag:
I think `default` is an important detail here, so either `default_area` or `default_dock_area` for extra clarity IMO

